PR #32, opened 2015-01-15
ARGO-390:  add is_file_in_workspace?

jmartin-sul:
make it easier to robustly determine whether a given object has a given file in its workspace.  add corresponding unit tests.


wmene:
Rebased commits on top of current develop

PR #39, opened 2015-01-28
Adding in a dedicated release node function and removing complexity for ...

carrickr:
...add_tag

@wmene 

Per our conversation yesterday about moving stuff to releasable


carrickr:
Three commits put into develop after this request make this pull rather ugly, closing.

PR #42, opened 2015-01-29
Logical ops

atz:
Some serious implosion of copy/paste in content_metadata_ds, otherwise mostly just logical operation cleanup.


atz:
Travis all green

PR #56, opened 2015-04-13
Rights indexing

atz:
Lots of cleanup and some purification in here.  Also, in various bits, initial dor-rights-auth integration work.


spec/dor/rights_metadata_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+    it "has a Dor::RightsAuth dra_object" do
+      expect(@rm.dra_object).to be_a(Dor::RightsAuth)
+      expect(@rm.dra_object.index_elements).to match a_hash_including(:primary => "stanford", :errors => [])
+    end
+  end
+
+  describe 'to_solr' do
+    it 'should have correct primary' do
+      # binding.pry
```
jmartin-sul: probably want to get rid of these binding.pry calls instead of leaving them in commented out?

atz: pulled in rebased version of subsequent patch

PR #58, opened 2015-04-20
Fix to correct the format of how tags that are deleted from purl are add...

carrickr:
...ed.

@atz @wmene 

Code fix for release tags.


carrickr:
@wmene @atz 

We should likely do this before develop diverges too far from the branch and rebase has to happen or gets messy?

wmene:
Not to be nit-picky, and I know it's only one line, but the fix you made is not exercised by any test (you'll see in the coverage report). I'll merge it in for now, but in the future, try to at least have a test that exercises what you change :-)

PR #64, opened 2015-06-13
Update presentable.rb

blalbrit:
adding in logo block. while not a requirement for IIIF spec compliance (this is an optional feature), this allows Mirador to call repository logo - allowing us to remove repo logos from the Mirador.


blalbrit:
re-file pull request per @wmene

PR #65, opened 2015-06-15
Iiif logo

blalbrit:
Adding static logo block to IIIF spec


blalbrit:
Re-filed pull request to appropriate branch per @wmene

PR #76, opened 2015-08-27
initial version of a republish scriptie that uses dor-services-app

drh-stanford:


drh-stanford:
this has been moved to dor-utils

PR #75, opened 2015-08-26
use https for logo URL

drh-stanford:
this is for the IIIF manifest's logo attribute

PR #78, opened 2015-09-01
use https for logo URL for 4.x

drh-stanford:
see https://github.com/sul-dlss/dor-services/pull/75 for discussion

### DIFF
--- a/lib/dor/models/presentable.rb
+++ b/lib/dor/models/presentable.rb
@@ -33,10 +33,10 @@ module Dor
         'label' => lbl,
         'attribution' => 'Provided by the Stanford University Libraries',
         'logo' => { 
-          '@id' => "http://stacks.stanford.edu/image/iiif/wy534zh7137%2FSULAIR_rosette/full/400,/0/default.jpg", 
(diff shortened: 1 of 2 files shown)


drh-stanford:
@atz @jmartin-sul are the builds broken on the 4.x branch? They are failing when trying to pull down gems... maybe it's spurious?

atz:
Build errors on old rubies for non-production dependency `debugger`.

atz:
After my revision of the Gemfile, can somebody determine why Travis (building under :test grouping, right?) is even trying to install debugger in the first place?  How is this dependency even being hit?

drh-stanford:
I was having problems with ruby versions. The debugger is keyed only to Ruby 1.9.x(?).

  https://github.com/sul-dlss/dor-services/blob/4.x/Gemfile#L7

atz:
Yeah, that's because debugger doesn't support ruby 2.x.  Most of these debugger/dev tools are still not developer dependencies, strictly speaking, and needn't be in our Gemspec.  In particular, stuff like `awesome_print` isn't even invoked anywhere, yet it is still required.  So I cleaned up spec_helper on 4.x.  Build is green there now, so rebase and PR away!

PR #79, opened 2015-09-02
add virtual resource to contentMetadata

drh-stanford:
See JIRA JUMBO-16 for issue tracking. This is related to https://github.com/sul-dlss/dor-utils/pull/4


drh-stanford:

drh-stanford:
Need to move to the 4.x branch.

PR #84, opened 2015-09-14
revise add_collection_reference to use location/url rather than ident…

drh-stanford:
…ifier to link collections. See LYBERSTRUCTURE-469 for details.


drh-stanford:

drh-stanford:
Need to move to the 4.x branch

PR #85, opened 2015-09-16
add support for externalFile references for IIIF manifest generation

drh-stanford:


drh-stanford:
@cbeer here's the implementation of virtual resources in the IIIF manifest

drh-stanford:
There's a problem with the public.xml in that it still have the verbatim contentMetadata from DOR, like so:

```xml
<contentMetadata objectId="hj097bm8879" type="map">
  <resource id="hj097bm8879_1" sequence="1" type="image">
    <externalFile fileId="2542A.jp2" objectId="druid:cg767mn6478" resourceId="cg767mn6478_1"/>
    <relationship objectId="druid:cg767mn6478" type="alsoAvailableAs"/>
  </resource>
  <resource id="hj097bm8879_2" sequence="2" thumb="yes" type="image">
    <externalFile fileId="2542B.jp2" objectId="druid:jw923xn5254" resourceId="jw923xn5254_1"/>
    <relationship objectId="druid:jw923xn5254" type="alsoAvailableAs"/>
  </resource>
</contentMetadata>
```

but things that consume PURL's public xml do not understand externalFile elements and thus would be ignored. A possible solution is to rewrite the contentMetadata (dereference the externalFile elements) before the public xml is saved to PURL.

LynnMcRae:
Basically we're viewing the public XML as the "raw" data, albeit enhanced in some cases, and was hoping we could do smooth stacks referencing universally though the manifest.  As Chris just mentioned, PURL still has a need to look directly at contentMetadata unfortunately, creating a second place to interpret <externalFile>.  We won't have a universal data types "manifest" for awhile, if ever (I think that's a desire across institutions tho).  I worry about publicXML bloat for objects with many many files though. I'll do a ticket for the PURL logic (maybe it's trivial in constructing a URL) and see if we need to consider alternatives.

drh-stanford:
I've changed the implementation so that the `ContentMetadataDS.public_xml` dereferences the `externalFile` links, and `build_iiif_manifest` uses only metadata from the public XML.

drh-stanford:

atz:
Probably need a rebase after the many merges/cherry-picks today.

drh-stanford:
Need to move this to the 4.x branch rather than develop.

PR #87, opened 2015-09-18
4.x: filter on MIME type image/jp2 for IIIF manifest files

drh-stanford:
see JUMBO-28


drh-stanford:
This needs to be ported to the 5.x branch when ready

atz:
cherry-picked to `develop`

PR #88, opened 2015-09-18
remove comments from MODS during publish step

drh-stanford:
See ARGO-401


drh-stanford:

atz:
Cherry-picked to `4.x`

PR #89, opened 2015-10-01
4.x: add support for externalFile references for IIIF manifest generation

drh-stanford:
This is a 4.x version of https://github.com/sul-dlss/dor-services/pull/85 from the 5.x branch. I deleted https://github.com/sul-dlss/dor-services/pull/85 in favor of this PR.


drh-stanford:
This PR is no longer valid... I need to do another PR for this issue.

PR #90, opened 2015-10-01
4.x: revise add_collection_reference to use location/url

drh-stanford:
rather than identifier type=uri. See LYBERSTRUCTURE-469 for details.

This is the 4.x version of https://github.com/sul-dlss/dor-services/pull/84 which is on 5.x. I deleted https://github.com/sul-dlss/dor-services/pull/84 in favor of this PR.

### DIFF
--- a/lib/dor/models/describable.rb
+++ b/lib/dor/models/describable.rb
@@ -121,16 +121,21 @@ module Dor
         title_node = Nokogiri::XML::Node.new('title', doc)
         title_node.content = collection_title
 
-        id_node = Nokogiri::XML::Node.new('identifier', doc)
-        id_node['type'] = 'uri'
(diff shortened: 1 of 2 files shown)


drh-stanford:
@atz here's another change for the publish workflow step... I'm doing my development off the 4.x branch now.

drh-stanford:
@atz, @LynnMcRae has asked whether this can go into production before we fire off our republish en masse. This means we need to cut a new 4.x release with this PR merged in. Thanks!

drh-stanford:
@LynnMcRae this PR is still pending review...

atz:
This looks find code-wise, but what are we doing to correct any data that was written under the old code, if anything?

atz:
Forward porting.

drh-stanford:
@LynnMcRae what did you want to do about all the existing objects that are published with the old-style MODS? Are we still planning a republish en-masse?

PR #91, opened 2015-10-01
4.x: add virtual resource to contentMetadata

drh-stanford:
This is the 4.x version of https://github.com/sul-dlss/dor-services/pull/79 which is on the 5.x branch. I closed https://github.com/sul-dlss/dor-services/pull/79 in favor of this PR.

See JUMBO-17 and the related PR sul-dlss/dor-utils#4

### DIFF
--- a/lib/dor/datastreams/content_metadata_ds.rb
+++ b/lib/dor/datastreams/content_metadata_ds.rb
@@ -67,10 +67,76 @@ module Dor
       save
     end
 
-    def add_resource(files,resource_name, position,type = "file")
-      xml = ng_xml
+    #
+    # Generates the XML tree for externalFile references. For example,
+    #
+    #     <externalFile objectId="druid:mn123pq4567" resourceId="Image01" fileId="image_01.jp2000" mimetype="image/jp2" />
+    #
+    # @param [String] objectId the linked druid
+    # @param [String] resourceId the linked druid's resource identifier
+    # @param [String] fileId the linked druid's resource's file identifier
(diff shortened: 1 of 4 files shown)


spec/dor/contentable_spec.rb
```diff
@@ -196,4 +196,27 @@ def internal_uri
       expect(obj.identityMetadata.tags).to include('Decommissioned : test')
     end
   end
+
+  describe '#add_constituent' do
+    let(:obj) do
+      o = Dor::Item.new
+      o
```
atz: This is silly.  Just do `Dor::Item.new` without the variable assignment.

lib/dor/datastreams/content_metadata_ds.rb
```diff
@@ -67,10 +67,77 @@ def add_file(file, resource_name)
...
+      # save the virtual resource as a sibling and return
+      self.ng_xml.root << resource
+      resource
+    end
+
+    def add_resource(files,resource_name, position,type="file") 
+      xml=ng_xml
       raise "resource #{resource_name} already exists" if xml.search('//resource[@id=\'' + resource_name + '\']').length > 0
-      max = -1
+      node=nil
```
atz: This assignment is meaningless and its existence is masked by block variables multiple times in this method, then reassigned later as well.

drh-stanford:
did a code cleanup to fix those issues

atz:
Ported to `develop`

PR #93, opened 2015-10-05
New use license types

jmartin-sul:
would appreciate review from @atz (in conjunction with the related argo changes)


atz:
Looks like you fixed the failing test in #a2d0080 but then they started failing again with subsequent commits?

atz:
Ah, I remember we figured this out.  It has something to do with a remote call getting cut off but is unrelated to the work in 70b293a.  Fakeweb is failing for Publishable.  Separate issue.

atz:
Ticketed Publishable issues at #94.  Note that previously passing tests now fail in the same way.

PR #95, opened 2015-10-12
Releaseable 4.x

lmcglohon:
Removed platform limitations on simplecov in Gemfile and removed requirement of pry in the spec_helper

PR #96, opened 2015-10-21
fix document_cache_host URLs to use https

jmartin-sul:


jmartin-sul:
@atz @lmcglohon 

so i was looking at laney's commits for PR #95 as a reference, but it turns out the changes here look like a subset of those.  maybe i should just look through those commits and cherry pick from them back to 5.x, and then close this?

jmartin-sul:
@atz pointed out that a lot of the other changes in PR #95 have already been made in 5.x, since rubocop has been run there.  so cherry picking those sounds like a mess of conflicts.

in somewhat poor form, i'll just merge this in myself and call it a day.

PR #98, opened 2015-10-29
keep Releaseable#to_solr from hitting PURL

jmartin-sul:
add an option to released_for that optionally omits the call to add_tags_from_purl.  default the new option to false (i.e. default to old behavior).  have to_solr call released_for(true) so that no request to PURL is made as part of solrization.

@lmcglohon @atz : is this something we'd want to backport to 4.x?

explanation for why this is being done:  lynn says we don't care about what actually lives on PURL when solrizing, just where we think the object should be released to according to its fedora data.  this fix allows solrization of objects that might have bad data in PURL (or which might live on instances without a corresponding PURL instance).  as such, it prevents some solrization errors, and when using this version of dor-services, argo's fixtures all load without a problem.  when argo's fixtures load without issue, argo's unit tests complete successfully.

### DIFF
--- a/lib/dor/models/releaseable.rb
+++ b/lib/dor/models/releaseable.rb
@@ -37,8 +37,9 @@ module Dor
     end
 
     # Determine which projects an item is released for
+    # @param [Boolean] skip_live_purl set true to skip requesting from purl backend
     # @return [Hash{String => Boolean}] all namespaces, keys are Project name Strings, values are Boolean
-    def released_for
+    def released_for(skip_live_purl = false)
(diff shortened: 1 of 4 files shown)


lmcglohon:
Not sure about the backporting to 4.x.

jmartin-sul:
i guess this info isn't part of what the 4.x branch indexes, since it was included in to_solr as part of the argo 3.x indexing overhaul.  since solrization was where the call to PURL was causing trouble, i think i'm inclined to not backport unless someone can think of a specific reason to.

PR #100, opened 2015-11-03
Add configuration of DOR Workflow Service log file location.

tingulfsen:


config/config_defaults.yml
```diff
@@ -61,5 +61,6 @@
 :workflow:
   :url:
   :timeout: 60
+  :logfile: workflow_service.log
```
drh-stanford: shouldn't this be `log/workflow_service.log`?

atz: In the context of this client, there is no `./log` directory.  In the context of your **application**, there is.

PR #101, opened 2015-11-03
add ruby 2.2.2 for compatibility with Argo

drh-stanford:
related to https://github.com/sul-dlss/argo/pull/90


drh-stanford:
I've added a reversion #105...

PR #103, opened 2015-11-03
Configure workflow service log with filename and rotation schedule.

tingulfsen:
dor-services-app uses branch 4.x of dor-services, so making this pull request also for branch 4.x.


atz:
Already PR'd to `develop` branch.

PR #104, opened 2015-11-03
Updating dor-workflow-service logging so that it also takes rotation …

tingulfsen:
…schedule into account.

Sorry - should have included this in the first pull request.


atz:
Committed separately to `4.x`

PR #106, opened 2015-11-03
removes the /legalcode or /us/legalcode in the creativeCommons URIs

drh-stanford:
Fixes part of https://github.com/sul-dlss/argo/issues/70

These data are hardcoded and probably should be migrated out into a YAML at some point.

### DIFF
--- a/lib/dor/models/editable.rb
+++ b/lib/dor/models/editable.rb
@@ -20,20 +20,20 @@ module Dor
     # out, since it will presumably use the same license code, but a different title and URI?
     CREATIVE_COMMONS_USE_LICENSES = {
       'by' =>       { :human_readable => 'Attribution 3.0 Unported',
-                      :uri => 'https://creativecommons.org/licenses/by/3.0/us/legalcode' },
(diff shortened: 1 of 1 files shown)


jmartin-sul:
re:  the `/legalcode` path in the URI:  i'd considered using the shorter version, but i figured it was safer to link directly to the full legal text instead of the summary?  i have no strong feelings on that, as i'm sure there are people who're much more authoritative on the matter than i am.

re:  moving that stuff to YAML.  i was about to say that i totally agree with that, but i'm not so sure, now that i think about it... is this really a setting?  it seems pretty fundamental to the content being edited, and it seems like something we want to keep a clear historical record of.  and i can see a case for tying this directly to the code:  it's not something that we want to differ from instance to instance, we just want it to change over time.  and we don't really want old entires to disappear completely (hence the deprecation mechanism).

mentioning @atz, who does have strong opinions on license URIs.

atz:
This at least does no harm and improves what we've got.  Agree w/ John that this is code, not configuration.  Would consider ultimately moving it to someplace more atomic (dor-rights-auth or some other place), but merging this for now.

PR #112, opened 2015-11-06
add ability to refresh cached version of workflows datastream

### LINKED ISSUE #111: workflow datastream should be in sync with workflow service content
The workflow data stream is being cached even when the workflow service's record of that data stream may have changed. Any calls that would write to the workflow service's record should invalidate the cache, at a minimum. 

Also, deprecate ``initialize_workflow` for `create_workflow` for consistency-sake

drh-stanford:
Fixes #111 

The `Dor::Item` object caches the XML for the workflows data stream. I've added the ability to refresh the cached XML from the workflow service using `item.workflows.content(true)`. I've also changed `create_workflow` (the new name for the deprecated `initialize_workflow`) so that after it calls the workflow service to create a new workflow, it refreshes the cache. I added unit tests for the cache invalidation logic.

@atz @jmartin-sul please review

### DIFF
--- a/lib/dor/datastreams/workflow_ds.rb
+++ b/lib/dor/datastreams/workflow_ds.rb
@@ -36,7 +36,11 @@ module Dor
       @ng_xml ||= Nokogiri::XML::Document.parse(content)
     end
 
-    def content
+    # @param [Boolean] refresh The WorkflowDS caches the content retrieved from the workflow
+    # service. This flag will invalidate the cached content and refetch it from the workflow
+    # service directly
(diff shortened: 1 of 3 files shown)


jmartin-sul:
:+1: from me.

if this gets merged, can you file an argo issue to switch it from using the deprecated `initialize_workflow` method name?

drh-stanford:
ok i added a todo comment to the argo PR for this fix. https://github.com/sul-dlss/argo/pull/113. go ahead and merge...

Issue #109, opened 2015-11-05
remove IIIF manifest generation from publish step


drh-stanford:
See https://github.com/sul-dlss/dor-services/blob/develop/lib/dor/models/publishable.rb#L66

For deployment, we need to:
- [x] Merge https://github.com/sul-dlss/dor-services/pull/128 for the public XML generation
- [x] Cut a 4.x dor-services release (`4.23.0`)
- [x] Deploy new dor-services 4.x into dor-services-app
- [x] Deploy new dor-services 4.x into common accessioning
- [ ] Clean up manifests in PURL document_store afterwards
- [x] Merge https://github.com/sul-dlss/dor-services/pull/157 for 5.x forward port of functionality

drh-stanford:
PR https://github.com/sul-dlss/dor-services/pull/114 is merged now.

atz:
`dor-services-app` now on `5.27.0`

jcoyne:
I don't think we know what we need to do here, so closing.

PR #114, opened 2015-11-10
4.x: remove IIIF manifest generation from publish step

drh-stanford:
Addresses #109. The publish robot no longer needs to create the IIIF manifest. I believe that no other application uses the IIIF manifest generation code, so it's deleted in this PR


atz:
Do we need to forward-port this to `5.x`?

atz:
Ported over.

PR #117, opened 2015-11-17
provide for setting empty copyright/uses and normalize defaultObjectRights XML

drh-stanford:
…ights XML. this includes the XML normalization code from the modsulator. 

Partially addresses Argo issue https://github.com/sul-dlss/argo/issues/70

### DIFF
--- a/dor-services.gemspec
+++ b/dor-services.gemspec
@@ -46,6 +46,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'dor-rights-auth', '~> 1.0', '>= 1.0.1'
   s.add_dependency 'lyber-utils', '~> 0.1.2'
   s.add_dependency 'moab-versioning', '~> 1.4.4' # 1.3.2 fails, 1.4.3 fails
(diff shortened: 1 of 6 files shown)


drh-stanford:
@tingulfsen please review this integration of the modsulator into dor-services. i'm using it on the defaultObjectRights datastream (in the rightsMetadata schema, not mods).

atz:
I'm not sure this can be cleanly picked over to `4.x`.  It seems to require several earlier commits including the `CREATIVE_COMMONS_USE_LICENSES` style variable declarations in Editable.  Even with those added, some 12 rspec errors are introduced with the pick.  Unless there is a specific `4.x` consumer that needs this, I think we should mark it as a point of divergence.

drh-stanford:
I think `develop` only is fine.

tingulfsen:
@drh-stanford sorry for the delay. I actually discovered an unrelated bug withing the MODSulator...arghhh...but I will get to the refactoring you require today and put forth a pull request for you to look at.

PR #122, opened 2015-11-21
use new modulator and nokogiri-pretty for XML normalization of defaultObjectRights

drh-stanford:
pending more tests...


drh-stanford:
@tingulfsen I've updated the tests for human readable output on a messy XML document. Please review.

PR #131, opened 2015-12-07
Registration service fixes

### LINKED ISSUE #127: register_object method duplicates solr writes
Calling `.save` and `.update_index` duplicates writes to solr in `register_object`:

https://github.com/sul-dlss/dor-services/blob/develop/lib/dor/services/registration_service.rb#L7

Also, at 100+ lines, that method is fugly, should be broken up.

### LINKED ISSUE #130: register_object test claims impossible/undesired resilience
Spec claimed that method should return object even when indexing fails.  But this only passed technically because it stubbed out `.save`, inside of which the **first** call to `.update_index` lives.  If that call were to fail, presumably under conditions identical to those that would cause the second call to fail, no error handling was provided, no object returned.  

The expectation should be that `.save` has passed without issue.  If indexing fails, it *should* raise exception.  In production for Argo, realtime indexing is disabled anyway!

atz:
Closes #130 
Closes #127

### DIFF
--- a/lib/dor/models/describable.rb
+++ b/lib/dor/models/describable.rb
@@ -41,31 +41,27 @@ module Dor
 
     # Generates Dublin Core from the MODS in the descMetadata datastream using the LoC mods2dc stylesheet
     #    Should not be used for the Fedora DC datastream
-    # @raise [Exception] Raises an Exception if the generated DC is empty or has no children
+    # @raise [CrosswalkError] Raises an Exception if the generated DC is empty or has no children
+    # @return [Nokogiri::Doc] the DublinCore XML document object
     def generate_dublin_core
-      format = metadata_format
-      if format.nil?
-        raise CrosswalkError, "Unknown descMetadata namespace: #{metadata_namespace.inspect}"
-      end
-      xslt = Nokogiri::XSLT(File.new(File.expand_path(File.dirname(__FILE__) + "/#{format}2dc.xslt")) )
+      raise CrosswalkError, "Unknown descMetadata namespace: #{metadata_namespace.inspect}" if metadata_format.nil?
+      xslt = Nokogiri::XSLT(File.new(File.expand_path(File.dirname(__FILE__) + "/#{metadata_format}2dc.xslt")) )
       desc_md = descMetadata.ng_xml.dup(1)
       add_collection_reference(desc_md)
       dc_doc = xslt.transform(desc_md)
-      # Remove empty nodes
-      dc_doc.xpath('/oai_dc:dc/*[count(text()) = 0]').remove
-      if dc_doc.root.nil? || dc_doc.root.children.size == 0
-        raise CrosswalkError, "Dor::Item#generate_dublin_core produced incorrect xml:\n#{dc_doc.to_xml}"
-      end
+      dc_doc.xpath('/oai_dc:dc/*[count(text()) = 0]').remove # Remove empty nodes
+      raise CrosswalkError, "Dor::Item#generate_dublin_core produced incorrect xml (no root):\n#{dc_doc.to_xml}" if dc_doc.root.nil?
+      raise CrosswalkError, "Dor::Item#generate_dublin_core produced incorrect xml (no children):\n#{dc_doc.to_xml}" if dc_doc.root.children.size == 0
       dc_doc
     end
 
+    # @return [String] Public descriptive medatada XML
     def generate_public_desc_md
       doc = descMetadata.ng_xml.dup(1)
       add_collection_reference(doc)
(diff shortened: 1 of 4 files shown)


lib/dor/models/describable.rb
```diff
@@ -41,31 +41,27 @@ def build_descMetadata_datastream(ds)
...
+    # @raise [CrosswalkError] Raises an Exception if the generated DC is empty or has no children
+    # @return [Nokogiri::Doc] the DublinCore XML document object
     def generate_dublin_core
-      format = metadata_format
-      if format.nil?
-        raise CrosswalkError, "Unknown descMetadata namespace: #{metadata_namespace.inspect}"
-      end
-      xslt = Nokogiri::XSLT(File.new(File.expand_path(File.dirname(__FILE__) + "/#{format}2dc.xslt")) )
+      raise CrosswalkError, "Unknown descMetadata namespace: #{metadata_namespace.inspect}" if metadata_format.nil?
+      xslt = Nokogiri::XSLT(File.new(File.expand_path(File.dirname(__FILE__) + "/#{metadata_format}2dc.xslt")) )
```
drh-stanford: We could put these xslt's into a cache rather than reparsing them everytime. There's really only 1 of these xslts `mods2dc.xslt`. It's overly flexible here.

lib/dor/services/registration_service.rb
```diff
@@ -4,6 +4,47 @@ module Dor
...
 
     class << self
+
+      # @TODO: Why isn't all this logic in, for example, Dor::Item.create? or Dor::Base.create? or Dor::Creatable.create?
+      # @TODO: these duplicate checks could be combined into 1 query
+
+      # @param [String] pid an ID to check, if desired.  If not passed (or nil), a new ID is minted
+      # @return [String] a pid you can use immidately, either freshly minted or your checked value
+      # @raise [Dor::DuplicateIdError]
+      def unduplicated_pid(pid = nil)
```
drh-stanford: This method name could be more clear. Maybe `check_or_mint_unique_pid` or something like that instead of `unduplicated_pid`? It's a little awkward that it's both a check function and a minting function.

lib/dor/services/registration_service.rb
```diff
@@ -18,23 +59,18 @@ def register_object(params = {})
...
-          existing_pid = SearchService.query_by_id(pid).first
-          unless existing_pid.nil?
-            raise Dor::DuplicateIdError.new(existing_pid), "An object with the PID #{pid} has already been registered."
-          end
-        else
-          pid = Dor::SuriService.mint_id
-        end
+
+        # Check for sourceId conflict *before* potentially minting PID
+        source_id_string = check_source_id [source_id.keys.first, source_id[source_id.keys.first]].compact.join(':')
```
drh-stanford: this resolution from source_id argument into a useful "abc:xyz" string is hard to read and uses source_id.keys.first twice -- maybe you should also do the fail `ArgumentError` before this line.

drh-stanford:
I have a few suggestions for the spec that I put here: [patch.txt](https://github.com/sul-dlss/dor-services/files/55967/patch.txt)

lib/dor/models/describable.rb
atz: Good point.  Not really related to my purely syntactical change though.

lib/dor/services/registration_service.rb
atz: Could rename it.  Not an uncommon pattern though, familiar from caching routines.  Don't care about the name much, but don't prefer the new one either.  If you think having the subroutine is bad, all I can say is that it *was* bundled in a single 100+ line method that did **everything** previously...

lib/dor/services/registration_service.rb
atz: The two calls to `keys.first` were in the existing code here:
https://github.com/sul-dlss/dor-services/blob/develop/lib/dor/services/registration_service.rb#L47

I just relocated it.  And I added one layer of ArgumentError checking here:
https://github.com/sul-dlss/dor-services/blob/registration_service_fixes/lib/dor/services/registration_service.rb#L73

Two calls to `keys` is costly only proportional to the size of the Hash.  In this case the max size of the Hash is now 1, so the cost is negligible.  Doesn't read great, but isn't my focus in this PR.

Ideally, the argument should be a single string since that is all the method ever wants to operate on (but that would be a breaking change, i.e. not in this PR).

lib/dor/models/describable.rb
atz: Ticketed in https://github.com/sul-dlss/dor-services/issues/135

PR #139, opened 2015-12-12
More rubocop compliance

atz:
Deprecated methods updated and a bunch of whitespace nits, mostly.


atz:
Review in this case is uninteresting.  Merging.

PR #142, opened 2015-12-14
index use_license_machine from rightsMetadata

drh-stanford:
and default_use_license_machine from defaultObjectRights.

This PR extracts the `use_license_machine` data from `rightsMetadata` and creates a new field for APOs called `default_use_license_machine` which is extracted from `defaultObjectRights`.

This PR also deletes some XML templates not being used in `RightsMetadataDS` and does some clarity cleanup for the specs.


drh-stanford:
not sure why coverage decreased...

PR #149, opened 2016-01-29
Enable simplecov to generate coverage data locally

darrenleeweber:


darrenleeweber:
The coverage may have decreased because some src code paths are excluded in the new config.

drh-stanford:
shouldn't this need an `add_development_dependency 'simplecov'` in the gemspec?

darrenleeweber:
`coveralls` already depends on `simplecov`, see
https://rubygems.org/gems/coveralls/versions/0.8.10

jmartin-sul:
unless it's pretty much unimaginable that coveralls would not depend on simplecov (as with something like Rails and ActiveWhatever), i'd be in favor of listing the dependency explicitly, if only for clarity, since this PR makes it a direct dependency.

FWIW, argo does the same thing, and lists simplecov as a direct dependency (which seems right to me):
https://github.com/sul-dlss/argo/blob/develop/Gemfile#L66

not to say that argo's a paragon of good design.  but it seems to me that a Gemfile should list exactly the direct dependencies of its application (and the indirect dependencies that require pinning due to whatever compatibility issues, e.g. as with so many of our applications and the Addressable gem for a while).

darrenleeweber:
OK, OK, so now it's explicit.

jmartin-sul:
:+1: as long as the tests pass, which i'm assuming they will.

question (maybe for @mejackreed, or someone else who's more well versed in ruby testing):  why is this a 6.6% drop in coverage?  i mean, i get that this should be non-zero drop in coverage, but the amount of code changed can't be over 6% of the codebase, right?  is this some artifact of more code being considered testable after this change?  something else entirely?

anyway, not saying we should hold this up over the coverage decrease, more that i'm perplexed by the reported size of the decrease, and am wondering if anyone has a better understanding of what's going on here.

mejackreed:
Seems like all of the drops came from adding the `gemfiles` filter https://github.com/sul-dlss/dor-services/pull/149/files#diff-93830fa29d616f7c87903d08b5b1b29aR7 

Do we need that filter?

darrenleeweber:
The idea is to let bundled gems test themselves.  I never want to see that stuff in the list of covered files.  Why should I look at coverage for a bundled gem?  If I need to know that stuff, I should be working on the gem directly, not indirectly.

jmartin-sul:
ah, ok.  that seems reasonable.  i'd think that percentage coverage reported should be for the app code itself, and that gem code shouldn't count towards the number.  same for spec code, unless you're testing the specs too (but then what will test the tests?  will the meta-tests count towards code coverage?).

anyway, :+1:

Issue #146, opened 2016-01-13
Review the 'groups_which_view_metadata'


darrenleeweber:
With regard to:
```
def can_view_metadata?(roles)
  intersect roles, groups_which_view_metadata
end
```

The `groups_which_view_metadata` is a misnomer, it should be `roles_which_view_metadata`.

With regard to related issues in argo, like https://github.com/sul-dlss/argo/issues/203 and https://github.com/sul-dlss/argo/issues/76, there may be users who belong to groups that have a role in the `dor-apo-viewer`, but this role will never be able to view an item because it is not included in the `roles_which_view_metadata`.  For example, in a spec of argo, we can see:
```
[11] pry(#<RSpec::ExampleGroups::Apo::WithViewingUser>)> dorObj.groups_which_view_metadata
=> ["dor-administrator", "sdr-administrator", "dor-apo-manager", "dor-apo-depositor", "dor-viewer", "sdr-viewer"]
[12] pry(#<RSpec::ExampleGroups::Apo::WithViewingUser>)> dorObj.roles
=> {"dor-apo-manager"=>["workgroup:dlss:developers"], "dor-apo-viewer"=>["person:sunetid:someone"]}
```

For this DOR APO object, the DOR object has a set of `groups_which_view_metadata` and a set of `roles`.  If the `dor-apo-viewer` role is matched for a user (based on the user workgroups), it doesn’t allow viewing the APO because there is no `dor-apo-viewer` in the `groups_which_view_metadata`.

So, for the APO `dorObj.groups_which_view_metadata`, should there be a `dor-apo-viewer`  in that set?

darrenleeweber:
Similarly,
```
> d = Dor::AdminPolicyObject.find(@druid)
=> #<Dor::AdminPolicyObject pid:"druid:ab987mn1234", >
> d.is_a? Dor::AdminPolicyObject
=> true
> d.can_view_content? ["dor-viewer"]
=> true
> d.can_view_content? ["dor-apo-viewer"]
=> false
> d.can_view_metadata? ["sdr-viewer"]
=> true
```

darrenleeweber:
This is a blocker for https://github.com/sul-dlss/argo/issues/76, where @LynnMcRae says the role definitions need to be updated, see https://github.com/sul-dlss/argo/issues/76#issuecomment-171129115

darrenleeweber:
[...]

PR #150, opened 2016-01-29
Revise roles permitted to act on APOs and Items.

darrenleeweber:
A permitted 'role' is distinguished from a work-'group'.  Argo has a
User#roles method that can translate from workgroups to roles, which is
the input to the `can*` public methods.  This also simplifies all these
methods by using constants for role names.  In case the role definitions
need to change, the private `role_which_*` methods can be modified as
necessary.  Some of these are alias_methods because they use the same
roles as the target method.  All of the `roles_which_*` methods should be
private, subordinate to the public `can*` methods.

Also removed deprecated roles for `dor-administrator` and `dor-viewer`

This implements the specs, at the time written, in this Consul document:
https://consul.stanford.edu/display/chimera/Repository+Roles+and+Permissions

This should fix for https://github.com/sul-dlss/dor-services/issues/146

### DIFF
--- a/lib/dor/models/admin_policy_object.rb
+++ b/lib/dor/models/admin_policy_object.rb
@@ -4,6 +4,7 @@ module Dor
     include Governable
     include Editable
     include Describable
+    include Permissable
     include Processable
     include Versionable
 
--- a/lib/dor/models/governable.rb
+++ b/lib/dor/models/governable.rb
@@ -27,7 +27,7 @@ module Dor
     end
 
     def reset_to_apo_default
-      rightsMetadata.content = admin_policy_object.rightsMetadata.ng_xml
+      rightsMetadata.content = admin_policy_object.rightsMetadata.ng_xml.to_s
     end
 
     def set_read_rights(rights)
@@ -52,14 +52,15 @@ module Dor
         when Dor::Collection
           collection_or_druid
       end
-
       collections.delete(collection)
       sets.delete(collection)
     end
+
     # set the rights metadata datastream to the content of the APO's default object rights
     def reapplyAdminPolicyObjectDefaults
       rightsMetadata.content = admin_policy_object.datastreams['defaultObjectRights'].content
     end
+
     def rights
       return nil unless self.respond_to? :rightsMetadata
       return nil if rightsMetadata.nil?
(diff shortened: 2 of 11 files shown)


lib/dor/models/governable.rb
```diff
@@ -75,56 +76,5 @@ def rights
         'None'
       end
     end
```
darrenleeweber: methods moved to permissable.

lib/dor/models/admin_policy_object.rb
```diff
@@ -4,6 +4,7 @@ class AdminPolicyObject < ::ActiveFedora::Base
     include Governable
     include Editable
     include Describable
+    include Permissable
     include Processable
     include Versionable
```
darrenleeweber: The includes for Permissable occur everywhere that Governable is included.

lib/dor/models/governable.rb
```diff
@@ -27,7 +27,7 @@ def default_workflow_lane
...
 
     def reset_to_apo_default
-      rightsMetadata.content = admin_policy_object.rightsMetadata.ng_xml
+      rightsMetadata.content = admin_policy_object.rightsMetadata.ng_xml.to_s
     end
```
darrenleeweber: Add a spec for this method and it failed without the `.to_s`.

lib/dor/models/permissable.rb
```diff
@@ -0,0 +1,160 @@
+module Dor
+  module Permissable
+    extend ActiveSupport::Concern
```
darrenleeweber: Not sure it needs to be an `ActiveSupport::Concern`.  Comments?

lib/dor/models/permissable.rb
```diff
@@ -0,0 +1,160 @@
+module Dor
```
darrenleeweber: Permissable implements https://consul.stanford.edu/display/chimera/Repository+Roles+and+Permissions

lib/dor/models/permissable.rb
```diff
@@ -0,0 +1,160 @@
...
+    # ---
+    # Common viewing permissions
+
+    def can_view_content?(roles)
+      intersect roles, roles_which_view_content
+    end
+
+    def can_view_metadata?(roles)
+      intersect roles, roles_which_view_metadata
+    end
```
darrenleeweber: The `private` methods can be modified to change role permissions without changing the public `can_*` methods.

spec/dor/governable_spec.rb
```diff
@@ -11,6 +11,12 @@ class GovernableItem < ActiveFedora::Base
   before(:each) { stub_config   }
   after(:each)  { unstub_config }
```
darrenleeweber: No substantial changes to existing specs (just moved some methods around).  Added one spec and moved others to permissable_spec.

spec/dor/permissable_spec.rb
```diff
@@ -0,0 +1,307 @@
+require 'spec_helper'
```
darrenleeweber: These specs implement the requirements in
https://consul.stanford.edu/display/chimera/Repository+Roles+and+Permissions

darrenleeweber:
The `alias_method` might need to be replaced with `alias`, per

darrenleeweber:
Moved to https://github.com/sul-dlss/dor-services/pull/183

PR #152, opened 2016-02-02
Stop truncating release_date in the model

atz:
Even if we wanted this, the model is the wrong place to do it.

And we don't want this.

### DIFF
--- a/lib/dor/datastreams/embargo_metadata_ds.rb
+++ b/lib/dor/datastreams/embargo_metadata_ds.rb
@@ -54,10 +54,10 @@ class EmbargoMetadataDS < ActiveFedora::OmDatastream
     term_values(:status).first
   end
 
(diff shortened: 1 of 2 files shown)


atz:
Obviously I don't care about a 0.0% "decrease" in coverage.

jmartin-sul:
:+1: 

do dor-services consumers have to worry about this?  e.g., are there any facet definitions that rely on the old truncating behavior?

i guess we don't have to worry about reindexing since the data being indexed was already truncated?

atz:
There may be one field where this was a non-date-comprehending approach to making a facet effectively "group by day".  If that is important, we can do the truncation *inside* `to_solr`.  

I would argue it is actually wrong, when we can use solr date math to accomplish the grouping dynamically.  (I may have already done that.)  In any case, the object shouldn't be truncating in memory when the truncation is for the index only.

PR #153, opened 2016-02-03
move basic DOR to Solr indexing code from Argo to dor-services (so other things can use it)

jmartin-sul:
code for indexing objects by ID (or list of IDs) is needed by many applications, and should be the same across them.  dor-services is a logical place to centralize this code, since the indexing code just ties together a few dor-services methods, with additional logging and fault tolerance options.

* move Argo::Indexer from the argo project to Dor::IndexingService in the dor-services project (except for reindex_pid_list_with_profiling, which will stay in argo, since it relies on the Argo::Profiler, which should not be moved to dor-services).
* move the corresponding tests to dor-services (except where related to the Argo::Indexer method that didn't get moved).
* spec_helper.rb: define a location for the indexing service log generated by running tests.
* config_defaults.yml: add config defaults for a file to which indexing events should be logged, and a format string for date entries in said indexing log.
* dor-services.rb: include the IndexingService class by default in the Dor module that dor-services consumers use.

### DIFF
--- a/spec/dor/indexing_service_spec.rb
+++ b/spec/dor/indexing_service_spec.rb
@@ -7,7 +7,6 @@ describe Dor::IndexingService do
 
   describe '#generate_index_logger' do
     before :each do
-      @mock_entry_id = 'unique_request_id'
       @mock_log_msg = 'something noteworthy'
     end
 
@@ -15,17 +14,30 @@ describe Dor::IndexingService do
       File.delete Dor::Config.indexing_svc.log if File.exist? Dor::Config.indexing_svc.log
     end
 
-    it 'should call entry_id_block and include the result in the logging statement' do
-      is_entry_id_block_executed = false
+    it 'should call entry_id_block for each logging event, and include the result in the logging statement' do
+      mock_req_id_ctr = 0
       test_index_logger = Dor::IndexingService.generate_index_logger do
-        is_entry_id_block_executed = true
-        @mock_entry_id
+        # this entry_id_block returns next value from a counter each time its called
+        mock_req_id_ctr += 1 # ruby has no [post-]increment operator
       end
-      test_index_logger.info @mock_log_msg
 
-      last_log_line = open(Dor::Config.indexing_svc.log).read.split("\n")[-1]
-      expect(last_log_line).to match(/\[#{@mock_entry_id}\] \[.*\] #{@mock_log_msg}$/)
-      expect(is_entry_id_block_executed).to eq(true)
+      # log some test messages
+      mock_log_messages = ['msg 1', 'msg 2', 'msg 3']
+      mock_log_messages.each do |msg|
+        test_index_logger.info msg
+      end
+
+      # parse the log and make sure the log lines match the expected format,
+      # with the expected messages, and the extra identifier (the counter
+      # values).
+      # the first log line is just a statement about when it was created, e.g.
+      # "# Logfile created on ...".  we just care about the subsequent lines.
+      log_lines = open(Dor::Config.indexing_svc.log).read.split("\n")[1..-1]
+      log_lines.each_with_index do |log_line, idx|
+        entry_id = idx + 1 # first entry_id is 1 since increment happened before return in entry_id_block
+        expect(log_line).to match(/\[#{entry_id}\] \[.*\] #{mock_log_messages[idx]}$/)
+      end
+      expect(mock_req_id_ctr).to eq(mock_log_messages.length)
     end
 
     it 'should log the default entry_id if entry_id_block is nil' do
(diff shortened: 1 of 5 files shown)


jmartin-sul:
TODO before moving out of WIP:
* remove the in-code TODO from the unit tests by either figuring out the problem, or figuring out that it can wait (and filing a bug)
* slight improvement to `IndexingService` logger machinery (and corresponding new tests):
 * trade `@@default_index_logger` for `@@loggers[:default]`.
 * then add `self.rails_index_logger`:
```
    def self.rails_index_logger
      @@loggers[:rails] ||= generate_index_logger { request.uuid }
    end
```

both argo and dor_indexing_app will want a Rails logger for indexing.  it's a pretty small amount of repeated code, but it's also pretty easy to do here, hence the second TODO.  plus, it'd save re-instantiating the logger on each request.

jmartin-sul:
also, haven't tested it yet, but i don't see anything obvious that'd keep us from just cherry picking this back to the 4.x branch.  which i believe means that dor_indexer_app's application code would be the same between v4 and v5, with the differences being in the:
* configuration (services pointed to)
* Gemfile and Gemfile.lock (dor-services version used)
* maybe the routes if `/v4/` vs `/v5/` is implemented there instead of at the webserver level.

`DorController.reindex` in dor_indexer_app should be like a simplified version of its counterpart in argo (it'd leave off the redirects and flash messages and only give a status code and message, since it's just being hit by other apps remotely, and not users in a more expansive web UI).

spec/dor/indexing_service_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+      expect(Dor).to receive(:load_instance).with(@mock_pid).and_raise(unexpected_err)
+      expect(@mock_default_logger).to receive(:warn).with(start_with("failed to update index for #{@mock_pid}, unexpected StandardError, see main app log: ["))
+      expect { Dor::IndexingService.reindex_pid(@mock_pid) }.to raise_error(unexpected_err)
+    end
+
+    it "should log the right thing if there's an unexpected Exception that's not StandardError, then re-raise the exception, even when should_raise_errors is false" do
+      stack_overflow_ex = SystemStackError.new "didn't see that one coming... maybe you shouldn't have self-referential collections?"
+      expect(Dor).to receive(:load_instance).with(@mock_pid).and_raise(stack_overflow_ex)
+      #TODO: file a bug for this expectation not working.  it seemed to work when this code was in argo, but doesn't now that it's been ported to dor-services.
+      # expect(@mock_default_logger).to receive(:error).with(start_with("failed to update index for #{@mock_pid}, unexpected Exception, see main app log: ["))
```
jmartin-sul: this is the only thing holding me from moving this out of WIP.  does anyone know why this might've worked in argo, but not here in dor-services?  same ruby version and all.  albeit, not rails.  but from playing with it in pry, it appears that `SystemStackError` can be trapped and logged in argo, but in these dor-services tests, the code path for the general `Exception` rescue that's used to log and rethrow the error never gets hit.

@mejackreed?  @atz?

jmartin-sul:
actually, one other change besides the testing TODO, see https://github.com/sul-dlss/argo/pull/496#issuecomment-179696789

spec/dor/indexing_service_spec.rb
mejackreed: A quick glance and that feels like a tricky one. Let's huddle up today and discuss? @atz ?

mejackreed:
Will there be a backport to the legacy dor-services version that argo 2.x is indexing off of?

jmartin-sul:
@mejackreed, re: backport:  yeah, that's my tentative plan, seems easy enough to cherry pick back to the dor-services 4.x branch.

jmartin-sul:
i think this should be ready for evaluation.

the first commit is close to a straight move of the `Argo::Indexer` functionality, as per the PR description (one method and its test stayed put in argo).  

there was an attempt to centralize generation of a logger for Rails, but that didn't work out as planned.  still, the test that was written for that had code that could improve an existing logging test, so the second commit is composed of the improved old test using code cannibalized from the defunct newer test.  plus, some comments about an unrelated minor regression (not a blocker, as per the comments).

atz:
I don't see the purpose of 4.x backport only because I don't think anybody who wants this feature (indexing) will be able to care about 4.x once the switch is made.

atz:
Leaving the branch here on the possibility that somebody will make a strong case for backporting.

jmartin-sul:
i wanted to backport to 4.x so that dor_indexing_app could use the same code to do indexing for its v4 incarnation.  then, the only difference between v4 and v5 of dor_indexing_app would be `Gemfile` and `Gemfile.lock` (and routes if that's where the version part of the URL gets set, though i'd rather do that at the webserver level unless ops objects to that as a poor plan).

PR #158, opened 2016-02-08
backport to 4.x: move DOR to Solr indexing code from Argo to dor-services (so other things can use it)

jmartin-sul:
NOTE:  this is a backport of https://github.com/sul-dlss/dor-services/pull/153, so that dor_indexing_app for v4 of dor-services can use the same indexing method signatures as it does with v5 of the dor-services gem.  below is the same summary given for the original 5.x implementation.

code for indexing objects by ID (or list of IDs) is needed by many applications, and should be the same across them.  dor-services is a logical place to centralize this code, since it just ties together a few dor-services methods, with additional logging and fault tolerance options.

* move Argo::Indexer from the argo project to Dor::IndexingService in the dor-services project (except for reindex_pid_list_with_profiling, which will stay in argo, since it relies on the Argo::Profiler, which should not be moved to dor-services).
* move the corresponding tests to dor-services (except where related to the Argo::Indexer method that didn't get moved).
* spec_helper.rb: define a location for the indexing service log generated by running tests.
* config_defaults.yml: add config defaults for a file to which indexing events should be logged, and a format string for date entries in said indexing log.
* dor-services.rb: include the IndexingService class by default in the Dor module that dor-services consumers use.

minor conflicts manually resolved in:
* lib/dor-services.rb: there'd been some formatting on the Services autoload section in dor-services v5, kept old style and added a line for IndexingService.
* spec/spec_helper.rb: 4.x had also added something to stub_config (stacks.document_cache_host).  kept that and added indexing_svc.log.

### DIFF
--- a/config/config_defaults.yml
+++ b/config/config_defaults.yml
@@ -65,3 +65,7 @@
   :shift_age: weekly
 :dor_services:
   :url:
+:indexing_svc:
+  :log: 'log/indexing_svc.log'
+  :log_date_format_str: '%Y-%m-%d %H:%M:%S.%L'
+  :log_rotation_interval: 'daily'
--- a/lib/dor-services.rb
+++ b/lib/dor-services.rb
@@ -132,6 +132,7 @@ module Dor
 
   # Services
   autoload :SearchService, 'dor/services/search_service'
+  autoload :IndexingService, 'dor/services/indexing_service'
   autoload :MetadataService, 'dor/services/metadata_service'
   autoload :RegistrationService, 'dor/services/registration_service'
   autoload :SuriService, 'dor/services/suri_service'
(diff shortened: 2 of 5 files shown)


atz:
In general no code should be written to use this (new indexing features on `4.x`).  New indexing features should all be on the newest major version.  So can we identify any potential consumers of this backport?

jmartin-sul:
dor_indexing_app for things that want to index using v4 of dor-services.  that's the only thing i can think of, but that's a big one, and seems worth it.  dor-services seems like a more logical home for this code than dor_indexing_app, since it was easy to cherry pick it to the relevant branch, and it keeps dor_indexing_app much simpler (and less divergent in its own codebase between the versions of dor-services it supports).

jmartin-sul:
i would not argue that we should backport this refactoring to argo 2.x, for example.  and i'd agree that things other than dor_indexing_app likely should not be calling this.

PR #159, opened 2016-02-11
upgrade stanford-mods dependency, update calling code and related tests

jmartin-sul:
* pin to stanford-mods 1.5.3, use 1.3.3
* pub_date_sort and pub_date_facet deprecated in favor of (respectively) pub_year_sort_str and pub_year_display_str.  update calls accordingly.
* index the newly available pub_year_int (an int equivalent of pub_year_sort_str) as sw_pub_date_sort_isi.
* update mocked pub_date_sort calls to pub_year_sort_str.
* add test info for new sw_pub_date_sort_isi field.
* require 'stanford-mods' instead of 'stanford-mods/searchworks', since the desired methods were moved to 'stanford-mods/origin_info' (but including just origin_info failed to load the method, and there seems to be no harm in loading the whole stanford-mods gem).

### DIFF
--- a/dor-services.gemspec
+++ b/dor-services.gemspec
@@ -47,7 +47,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'lyber-utils', '~> 0.1.2'
   s.add_dependency 'moab-versioning', '~> 1.4.4' # 1.3.2 fails, 1.4.3 fails
   s.add_dependency 'modsulator', '~> 0.0.7'
-  s.add_dependency 'stanford-mods', '= 1.3.3' # 1.3.4 fails on pub_date_sort, 1.4.0 fails
+  s.add_dependency 'stanford-mods', '= 1.5.3'  # v1.5.4 changes what sw_genre returns, which breaks test expectations.  TODO: figure out how to keep argo indexed counterparts of searchworks fields in sync w/ current SW behavior.
 
   # Bundler will install these gems too if you've checked out dor-services source from git and run 'bundle install'
   # It will not add these as dependencies if you require dor-services for other projects
--- a/lib/dor/models/describable.rb
+++ b/lib/dor/models/describable.rb
@@ -12,7 +12,7 @@ module Dor
       has_metadata :name => 'descMetadata', :type => Dor::DescMetadataDS, :label => 'Descriptive Metadata', :control_group => 'M'
     end
 
-    require 'stanford-mods/searchworks'
+    require 'stanford-mods'
 
     # intended for read-access, "as SearchWorks would see it", mostly for to_solr()
     # @param [Nokogiri::XML::Document] content Nokogiri descMetadata document (overriding internal data)
@@ -241,8 +241,9 @@ module Dor
           solr_doc[solr_key].push *vals unless vals.nil? || vals.empty?
           # asterisk to avoid multi-dimensional array: push values, not the array
         end
-        solr_doc['sw_pub_date_sort_ssi' ] = mods.pub_date_sort  # e.g. '0800'
-        solr_doc['sw_pub_date_facet_ssi'] = mods.pub_date_facet # e.g. '9th century'
+        solr_doc['sw_pub_date_sort_ssi' ] = mods.pub_year_sort_str  # e.g. '0800'
+        solr_doc['sw_pub_date_sort_isi' ] = mods.pub_year_int  # e.g. '0800'
+        solr_doc['sw_pub_date_facet_ssi'] = mods.pub_year_display_str # e.g. '9th century'
       end
       # some fields get explicit "(none)" placeholder values, mostly for faceting
       %w(sw_language_tesim sw_genre_tesim sw_format_tesim).each { |key| solr_doc[key] = ['(none)'] if solr_doc[key].empty? }
--- a/spec/dor/describable_spec.rb
+++ b/spec/dor/describable_spec.rb
@@ -573,7 +573,7 @@ describe Dor::Describable do
       expect{sm = @obj.stanford_mods}.not_to raise_error
       expect(sm).to be_kind_of(Stanford::Mods::Record)
       expect(sm.format_main).to eq(['Book'])
-      expect(sm.pub_date_sort).to eq('1911')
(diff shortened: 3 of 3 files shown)


dor-services.gemspec
```diff
@@ -47,7 +47,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'lyber-utils', '~> 0.1.2'
   s.add_dependency 'moab-versioning', '~> 1.4.4' # 1.3.2 fails, 1.4.3 fails
   s.add_dependency 'modsulator', '~> 0.0.7'
-  s.add_dependency 'stanford-mods', '= 1.3.3' # 1.3.4 fails on pub_date_sort, 1.4.0 fails
+  s.add_dependency 'stanford-mods' , '~> 1.5.3'
```
jmartin-sul: i figured it was reasonable to do *some* pinning, since the inadvertent move from 1.3.3 to 1.5.x introduced breakage.

lib/dor/models/describable.rb
```diff
@@ -12,7 +12,7 @@ class CrosswalkError < Exception; end
       has_metadata :name => 'descMetadata', :type => Dor::DescMetadataDS, :label => 'Descriptive Metadata', :control_group => 'M'
     end
 
-    require 'stanford-mods/searchworks'
+    require 'stanford-mods'
```
jmartin-sul: including `stanford-mods/origin_info` (where `pub_year_sort_str` and company live) didn't seem to resolve the method not found issue, but including the whole gem did.  any reason not to?  looks like @atz added that line, so seems like he'd be a good person to object if needed?  https://github.com/sul-dlss/dor-services/commit/4f4d871f20de2b2b51d4024b924752fedb468d21#diff-b031a0a86144a644f8641faccf39e94fR15

lib/dor/models/describable.rb
```diff
@@ -241,8 +241,9 @@ def to_solr(solr_doc = {}, *args)
           solr_doc[solr_key].push *vals unless vals.nil? || vals.empty?
           # asterisk to avoid multi-dimensional array: push values, not the array
         end
-        solr_doc['sw_pub_date_sort_ssi' ] = mods.pub_date_sort  # e.g. '0800'
-        solr_doc['sw_pub_date_facet_ssi'] = mods.pub_date_facet # e.g. '9th century'
+        solr_doc['sw_pub_date_sort_ssi' ] = mods.pub_year_sort_str  # e.g. '0800'
+        solr_doc['sw_pub_date_sort_isi' ] = mods.pub_year_int  # e.g. '0800'
+        solr_doc['sw_pub_date_facet_ssi'] = mods.pub_year_display_str # e.g. '9th century'
```
jmartin-sul: i would be open to renaming the solr field names to match the new stanford-mods method names, and updating references in argo.  opinions on that?

jmartin-sul:
opinions from @atz and @ndushay?

dor-services.gemspec
```diff
@@ -47,7 +47,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'lyber-utils', '~> 0.1.2'
   s.add_dependency 'moab-versioning', '~> 1.4.4' # 1.3.2 fails, 1.4.3 fails
   s.add_dependency 'modsulator', '~> 0.0.7'
-  s.add_dependency 'stanford-mods', '= 1.3.3' # 1.3.4 fails on pub_date_sort, 1.4.0 fails
+  s.add_dependency 'stanford-mods', '~> 1.5.3'
```
jmartin-sul: i figured it was reasonable to do some pinning, since the inadvertent move from 1.3.3 to 1.5.x introduced breakage.

jmartin-sul:
@ndushay stopped by for code review, says :+1:

jmartin-sul:
weird.  tests failing after rebase... investigating...

jmartin-sul:
turns out that 1.5.4 of stanford-mods had just been released, and changed the way sw_genre worked, causing expectations in dor-services tests to fail.  pinning to stanford-mods 1.5.3, will discuss how to properly stay up to date w/ searchworks fields next week (w/ @LynnMcRae, @ndushay, @lmcglohon, @atz).

ndushay:
MODS -> SearchWorks Solr index mappings are here:  
  https://github.com/sul-dlss/sw-indexer-service/blob/master/lib/sw_mapper.rb

pub year for sorting:  SearchWorks currently still uses a string (method  pub_year_sort_str) but will be migrating to an int value  (method   pub_year_int)

The date slider used by SearchWorks is currently populated with only a single value for each MODS record, using method  pub_year_int 

New Solr field pub_year_ss  populated by method  pub_year_for_display  is used for the value of the year in the title of each search result.   So 17uu  becomes 18th century, and 195x  becomes 1950s and other fun goodies.

jmartin-sul:
thanks!

dor-services.gemspec
```diff
@@ -47,7 +47,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'lyber-utils', '~> 0.1.2'
   s.add_dependency 'moab-versioning', '~> 1.4.4' # 1.3.2 fails, 1.4.3 fails
   s.add_dependency 'modsulator', '~> 0.0.7'
-  s.add_dependency 'stanford-mods', '= 1.3.3' # 1.3.4 fails on pub_date_sort, 1.4.0 fails
+  s.add_dependency 'stanford-mods', '= 1.5.3'  # v1.5.4 changes what sw_genre returns, which breaks test expectations.  TODO: figure out how to keep argo indexed counterparts of searchworks fields in sync w/ current SW behavior.
```
ndushay: the SearchWorks sprint is indeed changing stuff, and we need to be better about our semantic versioning.  I've goofed it up (1.3.4 should have been 1.4.0), and I forgot to tell @lmcglohon  to be sure to follow semantic versioning, too.  Sorry!   

the sw_genre values have changed;   Glancing thru the closed PRs or the recent commits is probably a fast way to figure out what happened.

lib/dor/models/describable.rb
ndushay: :+1:

PR #160, opened 2016-02-11
Update to moab-versioning 2.x, using fully namespaced Moab classes and modules

cbeer:


drh-stanford:
The builds are failing due to a ruby 2.1 constraint. Our Travis builds are doing 1.9.3, 2.0, 2.1, 2.2.

```
Gem::InstallError: moab-versioning requires Ruby version ~> 2.1.
```

cbeer:
Do we know who the Ruby 1.9 and 2.0 consumers are? I assumed we were addressing them with the grand argo roll-out?

jmartin-sul:
looking at the projects listed in the dor-services dependencies ticket, it seems at first glance like the following support ruby 1.9:
https://github.com/sul-dlss/assembly
https://github.com/sul-dlss/robot-suite-base
https://github.com/sul-dlss/dor-services-app
https://github.com/sul-dlss/hydrus/blob/master/.ruby-version

from what @atz was saying the other day, it seems like hydrus won't make the transition to dor-services 5.  and it sounded in the meeting this morning like  the plan was to start using the current ruby version for dor-services-app.  what about the other two?  is the intention to upgrade those?  i'd be in favor of doing that and dropping ruby 1.9 support in dor-services, if feasible.

cbeer:
Assembly is running under travis exclusively with ruby 2.1. I'll get a new version of ruby available on the robots VMs (and try to determine if there are any we can clean up...)

PR #163, opened 2016-02-17
Update Dor::Config to provide appropriate rest clients for SDR and DOR service applications

cbeer:


lib/dor/models/publishable.rb
```diff
@@ -69,9 +69,10 @@ def publish_metadata
 
     # Call dor services app to have it publish the metadata
     def publish_metadata_remotely
-      dor_services = RestClient::Resource.new(Config.dor_services.url + "/v1/objects/#{pid}/publish")
-      dor_services.post ''
-      dor_services.url
+      dor_services = Dor::Config.dor_services.rest_client
+      endpoint = dor_services["v1/objects/#{pid}/publish"]
+      endpoint.post ''
+      endpoint.url
```
darrenleeweber: It might help to add commentary on why the path construct is appended like this, because this construct may depend on the HTTP client library.  In the code removed, it's clear that `RestClient` is the library in use.  The new code only implies this by using `Config.dor_services.rest_client`.

lib/dor/config.rb
```diff
@@ -75,8 +75,11 @@ def make_solr_connection(add_opts = {})
             end
           }
         },
+        :dor_services => {
+          :rest_client => Confstruct.deferred { |c| config.make_rest_client c.url, c.cert_file, c.key_file, c.key_pass }
+        },
         :sdr => {
-          :rest_client => Confstruct.deferred { |c| config.make_rest_client c.url }
+          :rest_client => Confstruct.deferred { |c| config.make_rest_client c.url, c.cert_file, c.key_file, c.key_pass }
         },
```
darrenleeweber: The use of `rest_client` and `make_rest_client` seems to imply that `RestClient` is used, but it might help to add comments that explicitly clarify the expected client to be used.  Also, the specs might have to check that the config is using `RestClient` so that if, or when, the client library is changed out, the specs break and it will be easy to find the code locations that depend specifically on `RestClient`.

spec/dor/publishable_spec.rb
```diff
@@ -345,7 +345,7 @@ class ItemizableItem < ActiveFedora::Base
   describe 'publish remotely' do
     before(:each) do
       Dor::Config.push! {|config| config.dor_services.url 'https://lyberservices-test.stanford.edu/dor'}
-      allow_any_instance_of(RestClient::Resource).to receive(:post)
+      stub_request(:any, 'https://lyberservices-test.stanford.edu/dor/v1/objects/druid:ab123cd4567/publish')
     end
```
darrenleeweber: Maybe the spec could expect that the client used is `RestClient`?  If it is changed, the specs will fail and point the way to changing any code that depends specifically the `RestClient`.

darrenleeweber:
Generally :+1: for this PR, although it could be better with clarification about the HTTP client used and where the code depends specifically on `RestClient`.  If the intention is to have abstractions that don't depend on `RestClient`, the current PR may not do the trick.  That might require using Faraday and checking for any downstream implications.

lib/dor/models/publishable.rb
cbeer: This approach seems to be conventional for our other REST consumers. I don't think there's anything special about this particular one.

cbeer:
"Abstractions that don't depend on RestClient" certainly wasn't my intent here, just blending in with existing practice. I can't speak to the original author's intentions, but clearly the variations between ruby REST clients is inescapable, and I don't think there's any additional confusion introduced by my proposed changes.

PR #164, opened 2016-02-18
Use the dor-services REST API gateway to SDR services

cbeer:
Depends on https://github.com/sul-dlss/dor-services-app/pull/18 being deployed.


drh-stanford:
:+1: did you want this merged to develop before you deploy dor-services-app with SDR?

cbeer:
No, let's get the application deployed first.

Issue #166, opened 2016-02-18
revive master branch


drh-stanford:
we need to revive the `master` branch. it should be the release branch for 5.x. currently, `develop` is the 5.x branch and `4.x` is the 4.x branch.

PR #167, opened 2016-02-18
Merge commits from master into develop

cbeer:
Part of #166.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -117,11 +117,6 @@ Metrics/ClassLength:
 Metrics/CyclomaticComplexity:
   Max: 24
 
-# Offense count: 60
-# Configuration parameters: AllowURI, URISchemes.
-Metrics/LineLength:
-  Max: 6121
-
 # Offense count: 89
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
--- a/lib/dor/datastreams/workflow_ds.rb
+++ b/lib/dor/datastreams/workflow_ds.rb
@@ -1,7 +1,6 @@
 module Dor
   # TODO: class docs
   class WorkflowDs < ActiveFedora::OmDatastream
-
     set_terminology do |t|
       t.root(:path => 'workflows')
       t.workflow {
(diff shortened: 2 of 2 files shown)


lib/dor/datastreams/workflow_ds.rb
```diff
@@ -24,7 +23,7 @@ def get_workflow(wf, repo = 'dor')
       Workflow::Document.new(xml.to_s)
     end
 
-    alias_method :[], :get_workflow
+    alias :[] :get_workflow
```
darrenleeweber: I'd like to understand this better.  I've read recently that one or the other is more problematic.  From my recent reading, I concluded that `alias_method` is the way to go, but I'm not entirely clear on why one or the other is better.  For example, http://stackoverflow.com/questions/4763121/should-i-use-alias-or-alias-method

.rubocop_todo.yml
```diff
@@ -117,11 +117,6 @@ Metrics/ClassLength:
 Metrics/CyclomaticComplexity:
   Max: 24
 
-# Offense count: 60
-# Configuration parameters: AllowURI, URISchemes.
-Metrics/LineLength:
-  Max: 6121
-
```
darrenleeweber: By removing this and running rubocop again, does it complain about long lines?  Have they been fixed?

lib/dor/datastreams/workflow_ds.rb
```diff
@@ -9,7 +8,7 @@ class WorkflowDs < ActiveFedora::OmDatastream
         t.process {
           t.name_(    :path => {:attribute => 'name'     }, :index_as => [:displayable, :not_searchable] )
           t.status(   :path => {:attribute => 'status'   }, :index_as => [:displayable, :not_searchable] )
-          t.timestamp(:path => {:attribute => 'datetime' }, :index_as => [:displayable, :not_searchable] ) # , :data_type => :date)
+          t.timestamp(:path => {:attribute => 'datetime' }, :index_as => [:displayable, :not_searchable] ) #, :data_type => :date)
```
darrenleeweber: Has a library change (ActiveFedora?) made the extra `:data_type` arg obsolete and/or unnecessary?  Why leave the comment?

lib/dor/datastreams/workflow_ds.rb
cbeer: I believe @atz used `alias`, as suggested by the ruby style guide + rubocop:

cbeer: That said, I wouldn't worry too much about it. I think we should try to remove the `[]` accessor entirely. It seems to be used in one test in `dor-services`, but i'll take a look in some other projects just to be sure.

.rubocop_todo.yml
cbeer: Yes, it complains about long lines. And indeed, in a future commit, someone should fix that. I'd really rather not muddle getting `master` and `develop` back in sync with additional changes.

I'd also be concerned more if `dor-services` had a clean bill of health from rubocop, but there are already 200 violations (and the vast majority are not line length complaints).

lib/dor/datastreams/workflow_ds.rb
cbeer: Who knows. From the history, @atz was merely addressing rubocop violations in this file, and not necessarily trying to understand previous practice.

darrenleeweber:
I'm concerned that this merge is backwards and should be `develop` into `master`.

cbeer:
It should, except `master` got ahead of `develop` by these commits. This is just trying to set our development and release practices right again by getting the branches back in sync.

PR #171, opened 2016-02-19
Rename solrizer config

cbeer:
Includes #170

### DIFF
--- a/lib/dor/config.rb
+++ b/lib/dor/config.rb
@@ -55,8 +55,10 @@ module Dor
     end
 
     def make_solr_connection(add_opts = {})
-      opts = Config.solrizer.opts.merge(add_opts).merge(
-        :url => Config.solrizer.url
+      c = Dor::Config.solr || Dor::Config.solrizer
+
+      opts = c.opts.merge(add_opts).merge(
+        :url => c.url
       )
       ::RSolr::Ext.connect(opts)
     end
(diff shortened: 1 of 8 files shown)


lib/dor/config.rb
```diff
@@ -55,8 +55,10 @@ def make_rest_client(url, cert = Config.ssl.cert_file, key = Config.ssl.key_file
     end
 
     def make_solr_connection(add_opts = {})
-      opts = Config.solrizer.opts.merge(add_opts).merge(
-        :url => Config.solrizer.url
+      c = Dor::Config.solr || Dor::Config.solrizer
```
jmartin-sul: might be nice to leave a comment somewhere about supporting the deprecated config namespace, and/or give users of it a deprecation warning.

spec/dor/assembleable_spec.rb
```diff
@@ -12,8 +12,7 @@ class AssembleableItem < ActiveFedora::Base
 
     Dor::Config.push! do |config|
       config.suri.mint_ids false
-      config.gsearch.url  'http://solr.edu/gsearch'
-      config.solrizer.url 'http://solr.edu/solrizer'
+      config.solr.url 'http://solr.edu/solrizer'
```
jmartin-sul: i propose `config.solr.url 'http://solr.edu/solr'`for consistency :)

spec/dor_config.rb
```diff
@@ -22,11 +22,7 @@
     catalog.url 'http://example.edu/catalog/mods'
   end
 
-  gsearch do
-    rest_url 'https://example.edu/gsearch/rest'
-    url      'https://example.edu/solr/gsearch'
-  end
-  solrizer.url     'https://example.edu/solr/solrizer'
+  solr.url     'https://example.edu/solr/solrizer'
```
jmartin-sul: same comment as the other example.  i mean not a big deal if you'd prefer not, but more consistent w/ real solr urls too.

cbeer: I noticed there are existing tests that expect `https://example.edu/solr/solrizer`, and it wouldn't surprise me if there's other HTTP mocking too. I think it's a great idea, but I'd certainly prefer to defer it to a different pull request.

jmartin-sul: ok, makes sense

lib/dor/config.rb
cbeer: :ok_hand: I've added a deprecation warning to the `after configure` callback.

PR #172, opened 2016-02-19
Use the released version of dor-rights-auth

cbeer:


jmartin-sul:
:+1: by me, but @atz has done more w/ dor-rights-auth in dor-services, so opinion from him?

PR #173, opened 2016-02-19
Remove gemfiles from the travis configuration

cbeer:


cbeer:
Guess this depends of #172 .

PR #176, opened 2016-02-20
Copy Dor::Config solrizer parameters to the solr key

cbeer:


drh-stanford:
This PR deprecates the `Dor::Config.solrizer` configuration in favor of `Dor::Config.solr`.

PR #183, opened 2016-02-26
Revise roles permitted to act on APOs and Items.

darrenleeweber:
A permitted 'role' is distinguished from a work-'group'.  Argo has a
User#roles method that can translate from workgroups to roles, which is
the input to the `can*` public methods.  This also simplifies all these
methods by using constants for role names.  In case the role definitions
need to change, the private `role_which` methods can be modified as
necessary.  Some of these are alias_methods because they use the same
roles as the target method.  All of the `roles_which_` methods should be
private, subordinate to the public `can` methods.

Also removed deprecated roles for `dor-administrator` and `dor-viewer`

This implements the specs, at the time written, in this Consul document:
https://consul.stanford.edu/display/chimera/Repository+Roles+and+Permissions

This should fix https://github.com/sul-dlss/dor-services/issues/146

### DIFF
--- a/dor-services.gemspec
+++ b/dor-services.gemspec
@@ -7,8 +7,8 @@ Gem::Specification.new do |s|
   s.name        = 'dor-services'
   s.version     = Dor::VERSION
   s.platform    = Gem::Platform::RUBY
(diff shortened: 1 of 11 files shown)


darrenleeweber:
The alias_method might need to be replaced with alias, per

LynnMcRae:
Just a clarification … users get into roles by being assigned directly by sunetid or by being in a workgroup that is assigned to a role. While only the latter requires fancy footwork to find the user via group membership, I assume User#roles covers both situations?

Lynn

PR #184, opened 2016-02-26
Use and store the instantiatable Dor::WorkflowService within the Dor:…

cbeer:
…:Config

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -26,6 +26,7 @@ require 'nokogiri'
 
 require 'dor_config'
 require 'vcr'
+require 'retries'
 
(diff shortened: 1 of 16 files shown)


atz:
This is good, is required to get away from the old singleton pattern.

spec/spec_helper.rb
```diff
@@ -26,6 +26,7 @@
 
 require 'dor_config'
 require 'vcr'
+require 'retries'
```
atz: Where is this newly used?

atz: OK, not new, but `spec_helper` later disables retry sleep for tests.  Reasonable to have here.

PR #185, opened 2016-02-26
Use the dor-services REST API gateway to SDR services

cbeer:


atz:
Ignoring coveralls **-0.03%** on this one.

Issue #180, opened 2016-02-23
Why is modsulator a dependency of dor-services?


cbeer:
I don't see where modsulator is used in the dor-services code; why is it bundled into dor-services?

LynnMcRae:
Modsulator turns rows and columns (CSV or XLS) into MODS XML.  It is just one step to getting a spreadsheet loaded into DOR.  Modsulator itself should have no DOR dependencies. This may be an artifact of the original offline version of the facility when the conversion was bound in with the loading?

cbeer:
Ok, it turns out `dor-services` is using `modsulator` for some XML normalization routines.

LynnMcRae:
are you sure it's modsulator? (a rhetorical question) ... There's an XML normalizer that modsulator uses that should be available separately.  See https://consul.stanford.edu/display/chimera/MODS+bulk+loading 

If they are intertwined, it would be good to separate them.

cbeer:
Yes, it's all wrapped up in modsulator's normalizer.

drh-stanford:
We're only using a subset of modsulator for normalizing the `rightsMetadata`. I think it'd be fine to create a common gem that only has the basic XML normalization functions. See https://github.com/sul-dlss/argo/issues/71

atz:
Resolved by #344

PR #186, opened 2016-02-26
Remove modsulator dependency

cbeer:
See https://github.com/sul-dlss/dor-services/issues/180

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -390,7 +390,7 @@ Style/SingleLineMethods:
 
 # Offense count: 76
 # Cop supports --auto-correct.
-Style/SingleSpaceBeforeFirstArg:
+Style/SpaceBeforeFirstArg:
   Exclude:
     - 'config/dev_console_env.rb'
     - 'lib/dor/datastreams/content_metadata_ds.rb'
--- a/dor-services.gemspec
+++ b/dor-services.gemspec
@@ -43,7 +43,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'dor-rights-auth', '~> 1.0', '>= 1.0.2'
   s.add_dependency 'lyber-utils', '~> 0.1.2'
   s.add_dependency 'moab-versioning', '~> 2.0'
-  s.add_dependency 'modsulator', '~> 0.0.7'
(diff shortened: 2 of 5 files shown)


cbeer:
Sorry, I think that context got lost when switching from `develop` to `master` (see #181)

> This copies the only the relevant normalization from modsulator into a Dor::Utils::Normalizer class. @tingulfsen and I discussed alternatives (including copying normalizer into a separate gem), but we agreed the similarity is relatively trivial and may not be worth doing.

The main reason to break the dependency is we really shouldn't drag in all of modsulator's dependencies (including some heavy-weight ones like `roo`) just to clean up some rights XML. It's especially confusing because Modsulator's Normalizer class has really been set up to handle MODS.

We certainly could make a brand new gem with just this class. I'd suggest we wait until we have some other consumer or find these classes are changing before bothering with the overhead. There's a separate question to discover what is producing bad rights metadata and just fix it.

drh-stanford:
@tingulfsen i agree with @atz and propose that we extract the pure XML normalization into a `Normalizer` gem (rather than calling it `Modsulator::Utils`) as it would be a place to put our normailzation code -- beyond MODS and RightsMetadata, which is what it covers now. Dor services already has lots of code bloat and it'd be nice to keep the duplicate code out.

Re: producting bad rights metadata -- it's users who can type in randomly weird XML into the rightsMetadata editor in Argo. The editor validates the user-entered XML but does not normalize it (it's in JavaScript -- see https://github.com/sul-dlss/argo/blob/master/app/assets/javascripts/modules/datastream_edit.js#L9)

LynnMcRae:
MODS Normalization was specified to be its own separate service (see https://consul.stanford.edu/display/chimera/MODS+bulk+loading) callable by Modsulator but able to be independently invoked, including a web service call. I don't know if we actually have that, but making it generic is a good thing, more valuable than a MODS-only normalizer, even if it retains some small extra knowledge of MODS.  I'd like to have the Normalization page on consul properly reflect the more generic capability. I'll work with @tingulfsen on that if he's tracking it.

ndushay:
@tingulfsen is this addressed by #209 ?

jcoyne:
Superseded by #344

PR #192, opened 2016-03-02
Stop indexing the workflow archive count

cbeer:
Calculating the count is an expensive operation, and the value is no longer
being presented downstream.

### DIFF
--- a/spec/dor/workflow_object_spec.rb
+++ b/spec/dor/workflow_object_spec.rb
@@ -22,10 +22,8 @@ describe Dor::WorkflowObject do
       @item.workflowDefinition.content = '<workflow-def id="accessionWF"/>'
     end
 
-    it 'indexes the number of archived objects for the workflow' do
-      expect(Dor::Config.workflow.client).to receive(:count_archived_for_workflow).and_return(5)
-      expect(@item.to_solr).to include 'accessionWF_archived_isi' => 5
+    it 'indexes the workflow name' do
+      expect(@item.to_solr).to include 'workflow_name_ssim' => 'accessionWF'
     end
   end
-
 end
--- a/spec/dor/workflow_object_spec.rb
+++ b/spec/dor/workflow_object_spec.rb
@@ -23,7 +23,7 @@ describe Dor::WorkflowObject do
     end
 
     it 'indexes the workflow name' do
-      expect(@item.to_solr).to include 'workflow_name_ssim' => 'accessionWF'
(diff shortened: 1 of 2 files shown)


LynnMcRae:
I'm wondering if the workflow-archiver is triggering reindexing of the WFs in order to force a recount, something it would no longer have to do.

cbeer:
I don't think it is -- looking in the apache logs, I didn't see any signs of regular indexing for those workflow objects.

LynnMcRae:
no wonder the count was never accurate!

jmartin-sul:
i just pushed a commit which should hopefully fix the failing test.  was a simple mismatch in field type expectations (though we defaulted to `_ssim` too often, and this might be one of the cases where a single valued `_ssi` would've been more appropriate).

spec/dor/workflow_object_spec.rb
```diff
@@ -22,10 +22,8 @@
       @item.workflowDefinition.content = '<workflow-def id="accessionWF"/>'
     end
 
-    it 'indexes the number of archived objects for the workflow' do
-      expect(Dor::Config.workflow.client).to receive(:count_archived_for_workflow).and_return(5)
-      expect(@item.to_solr).to include 'accessionWF_archived_isi' => 5
+    it 'indexes the workflow name' do
+      expect(@item.to_solr).to include 'workflow_name_ssim' => ['accessionWF']
     end
   end
```
darrenleeweber: It seems the spec on the `to_solr` method did not previously check for the name, but it was already created by some other part of the `item.to_solr` machinery.  These diffs suggest that removing the `to_solr` method on the workflow object would cause this spec to fail with a method missing error.

darrenleeweber:
Seems harmless enough, so :+1: but defer to @atz for insider knowledge on Solr issues and implications.

spec/dor/workflow_object_spec.rb
cbeer: Indeed. I added this test to provide an equal level of confidence in our workflow object indexing as we had before.

jmartin-sul: i believe the parent provides a to_solr method (the original calls `super solr_doc, *args`).  i think essentially all dor-services objects have to_solr methods, courtesy of... `ActiveFedora::Base`?  

the tests pass with this call (and weren't previously failing w/ a method missing error, it was a mismatch in actual vs expected field value).

PR #193, opened 2016-03-02
fix missing variable name in get_content_diff

drh-stanford:


darrenleeweber:
This is exactly the kind of thing that will CHURN because we are not consistent across DOR and SDR in the use of the term DRUID.  I believe PID is a Fedora thing.  If that's right, I would pitch an argument for purging that term completely from everything in DOR/Argo.  If it's a term closely tied to a persistence framework, isolation from it is better.  The predominant term for any object identifier in DOR should be DRUID.  The only term used consistently in SDR is already DRUID.  Whether or not teams agree on the outcome of debates on this term will be reflected directly in changes like this one.

So, I don't want to blow up what looks like a trivial PR into some huge political issue.  I just want to make a passing point about the inconsistency in the system and how it might lead to fruitless churn on using one or another term.

drh-stanford:
This PR is a fix because `druid` is not a defined variable at that line of the code. Sorry I didn't state that up front (we were discussing it on slack). So, `pid` is used instead in the fix as that's a defined attribute from `ActiveFedora::Base`.

Re: naming. I agree that we should be calling these uuid values druids. I don't know what value a "pid" name has for us.

darrenleeweber:
Assuming this passes specs (how did it pass before?), then move right along and merge it.

drh-stanford:
As far as I can tell, this specific method is untested. Itemizable has basically no coverage.

drh-stanford:
I've added a simple test for `get_content_diff` that would have caught the undefined variable exception.

PR #194, opened 2016-03-03
Have Dor.configure initialize the Dor::WorkflowService singleton

cbeer:
This fixes a regression for clients that assumed `dor-services` would automatically configure `Dor::WorkflowService`.

### DIFF
--- a/lib/dor/config.rb
+++ b/lib/dor/config.rb
@@ -25,6 +25,14 @@ module Dor
       ensure
         $-v = temp_v
       end
+
+      # Initialize and configure the global Dor::WorkflowService for clients that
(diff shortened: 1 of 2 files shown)


cbeer:
:shrug: I'm slightly inclined for us to just close this and fix up badly behaved robots; on the other hand, it does no harm and allows us to fix such defects in our own time.

drh-stanford:
Yeah I'm inclined to keep updating the consumer code rather than this PR. 

The design is to use this lazy eval: https://github.com/sul-dlss/dor-services/blob/wfs-singleton/lib/dor/config.rb#L92

I have the config/environments workaround for now until we can deploy those PRs above.

PR #197, opened 2016-03-10
get_content_group_diff should handle the case when contentMetadata is missing

drh-stanford:
This PR fixes https://github.com/sul-dlss/argo/issues/651 by returning an empty `FileGroupDifference` when `contentMetadata` is missing.

### DIFF
--- a/lib/dor/models/itemizable.rb
+++ b/lib/dor/models/itemizable.rb
@@ -23,20 +23,19 @@ module Dor
 
     # Retrieves file difference manifest for contentMetadata from SDR
     #
-    # @param [String] subset keyword for file attributes :shelve, :preserve, :publish. Default is :all.
+    # @param [Symbol] subset keyword for file attributes :shelve, :preserve, :publish. Default is :all.
     # @param [String] version
-    # @return [String] XML contents of cm_inv_diff manifest
+    # @return [Moab::FileInventoryDifference] XML contents of cm_inv_diff manifest
     def get_content_diff(subset = :all, version = nil)
       if Dor::Config.stacks.local_workspace_root.nil?
         raise Dor::ParameterError, 'Missing Dor::Config.stacks.local_workspace_root'
       end
 
-      current_content = datastreams['contentMetadata'].content
-      if current_content.nil?
+      if contentMetadata.nil?
         raise Dor::Exception, 'Missing contentMetadata datastream'
       end
 
-      Sdr::Client.get_content_diff(pid, current_content, subset, version)
+      Sdr::Client.get_content_diff(pid, contentMetadata.content, subset, version)
     end
   end
 end
--- a/lib/dor/services/technical_metadata_service.rb
+++ b/lib/dor/services/technical_metadata_service.rb
@@ -50,6 +50,8 @@ module Dor
     def self.get_content_group_diff(dor_item)
       inventory_diff = dor_item.get_content_diff('all')
       inventory_diff.group_difference('content')
+    rescue Dor::Exception # no contentMetadata
+      Moab::FileGroupDifference.new
     end
 
     # @param [FileGroupDifference] content_group_diff
--- a/lib/dor/utils/sdr_client.rb
+++ b/lib/dor/utils/sdr_client.rb
@@ -33,6 +33,7 @@ module Sdr
         Moab::SignatureCatalog.new(:digital_object_id => druid, :version_id => 0)
       end
 
+      # @return [Moab::FileInventoryDifference] the differences for the given content and subset
       def get_content_diff(druid, current_content, subset = :all, version = nil)
         unless %w(all shelve preserve publish).include?(subset.to_s)
           raise Dor::ParameterError, "Invalid subset value: #{subset}"
--- a/spec/dor/itemizable_spec.rb
+++ b/spec/dor/itemizable_spec.rb
@@ -11,17 +11,21 @@ describe Dor::Itemizable do
   after(:each)  { unstub_config }
 
   before :each do
-    @item = instantiate_fixture('druid:ab123cd4567', ItemizableItem)
-    @item.contentMetadata.content = '<contentMetadata/>'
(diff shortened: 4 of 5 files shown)


lib/dor/models/itemizable.rb
```diff
@@ -23,20 +23,19 @@ def clear_diff_cache
...
-      if current_content.nil?
-        raise Dor::Exception, 'Missing contentMetadata datastream'
-      end
+      # If we don't have a contentMetadata datastream, return an empty difference.
+      # The technical-metadata robots relies on this functionality.
+      return Moab::FileInventoryDifference.new if datastreams['contentMetadata'].nil?
 
-      Sdr::Client.get_content_diff(pid, current_content, subset, version)
+      Sdr::Client.get_content_diff(pid, datastreams['contentMetadata'].content, subset, version)
     end
```
darrenleeweber: Why can't the technical-metadata robots just handle the `Dor::Exception` and move along without changing this code?

drh-stanford: Yeah, I prefer not to change the code here as I haven't pinpointed the source of the behavior change. The technical-metadata robot doesn't do anything other than build the technicalMetadata datastream here (in the abstract shared version):  https://github.com/sul-dlss/common-accessioning/blob/master/robots/accession/abstract_metadata.rb#L21 that code hasn't changed in  a couple years.

darrenleeweber: Perhaps the change needs to be in https://github.com/sul-dlss/dor-services/blob/master/lib/dor/services/technical_metadata_service.rb#L48-L53 or somewhere near there?  It may need to have some differentiation between content vs. non-content objects.

drh-stanford: Yeah, it seems like all this code here: https://github.com/sul-dlss/dor-services/blob/master/lib/dor/services/technical_metadata_service.rb#L14-L28 is only for when an item has contentMetadata.

drh-stanford: here's a possible example for catching that exception -- is that what you had in mind?
```
    def self.get_content_group_diff(dor_item)
      inventory_diff = begin
          dor_item.get_content_diff('all')
        rescue Dor::Exception # no contentMetadata
          FileInventoryDifference.new
        end
      inventory_diff.group_difference('content')
    end
```

darrenleeweber: Yes, that looks like a reasonable exception handling pattern.  It would help if the `Dor::Exception` either had a more specific sub-class or a useful message that could confirm the reason for the exception is no content-md.  The exception handler might need to confirm that the object is a non-content-md before carte-blanche creating an empty `FileInventoryDifference`.  If it is not a non-content-md object, then the exception might have to be raised again?

ndushay:
@jmartin-sul @tingulfsen   this PR is to fix a must-have bug (#651 - Technical Metadata requires contentMD for non-content-bearing objects)  ...  can it be merged and deployed to dev or stage for testing?

lib/dor/models/itemizable.rb
```diff
@@ -23,20 +23,19 @@ def clear_diff_cache
...
       end
 
-      current_content = datastreams['contentMetadata'].content
-      if current_content.nil?
+      if contentMetadata.nil?
         raise Dor::Exception, 'Missing contentMetadata datastream'
       end
 
-      Sdr::Client.get_content_diff(pid, current_content, subset, version)
+      Sdr::Client.get_content_diff(pid, contentMetadata.content, subset, version)
```
ndushay: it looks like before this PR, we checked:
- ```datastreams['contentMetadata'].content```  for nil

and now we check
- ```contentMetadata```  for nil

Questions:
- is ```contentMetadata``` the same as  ```datastreams['contentMetadata']```
- shouldn't we check BOTH  ```datastreams['contentMetadata']```  AND  ```datastreams['contentMetadata'].content``` for nil?

jmartin-sul: IIRC:
* `contentMetadata` is the same as `datastreams['contentMetadata']`
* if `contentMetadata` exists, `contentMetadata.content` might be empty string, but it won't be `nil`.

i've not tested either of these things.  maybe @atz can chime in, as more of an activefedora expert than me?

lib/dor/services/technical_metadata_service.rb
```diff
@@ -50,6 +50,8 @@ def self.test_jhove_service
     def self.get_content_group_diff(dor_item)
       inventory_diff = dor_item.get_content_diff('all')
       inventory_diff.group_difference('content')
+    rescue Dor::Exception # no contentMetadata
+      Moab::FileGroupDifference.new
     end
```
jmartin-sul: is this the most specific exception that can be caught here?

jmartin-sul:
this looks ok to me, but i'm probably not the person best qualified to judge this PR

lib/dor/services/technical_metadata_service.rb
drh-stanford: `get_content_diff` throws Dor::Exception or Dor::ParameterError directly -- only the Dor::Exception case is of interest. The SdrClient could throw other exceptions but those probably should throw upwards.

lib/dor/models/itemizable.rb
drh-stanford: The `item.contentMetadata` is the accessor to the datastream. See here for where it gets established: https://github.com/sul-dlss/dor-services/blob/182f59f0cb66577eef3a8bd8a5aaeea3265b2ef1/lib/dor/models/itemizable.rb#L9

The `.content` yields the XML string for the data stream. It's an ActiveFedora method. You can check for it to be `.blank?` but not `.nil?` I think.

ndushay:
fyi, also hoping this will solve sul-dlss/argo#697  (shelve robot fails for same reason)

ndushay:
@atz   you have been identified as the desired person to review this PR>

PR #201, opened 2016-03-11
Fix collection reference 404

jmartin-sul:
refactoring in small-ish increments, as seen in the commit history, then bug fixed in the last commit.

this assumes that we want to create a `relatedItem` node even if the reference is to a non-existent collection (in which case this change just stored the druid where it would store the collection name, a la the indexing behavior).  if we just want to not create a `relatedItem` node in these situations, that's easy enough also.  @LynnMcRae?

### DIFF
--- a/lib/dor/models/describable.rb
+++ b/lib/dor/models/describable.rb
@@ -101,6 +101,13 @@ module Dor
       end
     end
 
+    # Remove existing relatedItem entries for collections from descMetadata
+    def remove_related_item_nodes_for_collections(doc)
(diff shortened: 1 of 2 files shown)


drh-stanford:
I'm :+1: for this PR. It's still marked [WIP] though...

LynnMcRae:
@jmartin-sul : a bad collection reference is bad data no matter how you slice it, so my first instinct was to skip creating a relatedItem in the MODS and not create a link to nowhere. On the other hand a Collection title that is a druid will be noticed and fixed vs absent data that might go unnoticed ... but I'm still with the first option (no relatedItem) if that is as easy as the other, given we're talking about public data. @blalbrit do you agree?

blalbrit:
+1 @LynnMcRae

jmartin-sul:
ok, will change to not write it in that case...

jmartin-sul:
fixed

PR #202, opened 2016-03-12
Rubocop fixes

ndushay:
The story behind these fixes:  I was examining the violations of style rules in dor-services to see examples of violations and whether or not I agreed with keeping the rule.   In most of these cases, it was quick to just go ahead and do the fix, so I did.

I understand that these may be annoyances for existing open PRs;  I tried to separate each fix and am happy to revise.   Or this can be closed, if you like.

@jmartin-sul


lib/dor/models/identifiable.rb
```diff
@@ -210,10 +210,11 @@ def update_tag(old_tag, new_tag)
...
       # desc_md_ds_title will be nil if related_obj is nil
-      desc_md_ds_title = if related_obj
-        desc_md_ds = related_obj.datastreams['descMetadata']
-        desc_md_ds ? desc_md_ds.title_info.main_title.first : nil
-      end
+      desc_md_ds_title =
+        if related_obj
+          desc_md_ds = related_obj.datastreams['descMetadata']
+          desc_md_ds ? desc_md_ds.title_info.main_title.first : nil
+        end
```
atz: This assignment is sub-optimal to begin with.  It could be simplified to include:
```ruby
return default_title unless related_obj
```

That removes the entire weird assignment conditional prompting this formatting issue.

ndushay: That would have been better, had I thought of it.

PR #204, opened 2016-03-15
Add relationship access for constituent_object

cbeer:
does this address #206?


atz:
There should probably be a test to go with this, right?

cbeer:
Replaced by #263.

PR #209, opened 2016-03-26
Remove nokogiri-pretty and replace it by our own method to prettify/i…

tingulfsen:
…ndent XML for human consumption.


tingulfsen:
This pull request is a partial fix for sul-dlss/argo#691. nokogiri-pretty no longer seems to be maintained, and the published nokogiri-pretty gem does not seem to match up with its current GitHub code. This pull request adds our own XSLT file and corresponding code to ensure that the stored XML is nicely indented AND declares the encoding to be UTF-8.

Issue #25, opened 2014-12-02
equivalent-xml dependency pegged at 3-year-old version


atz:
The **dor-services.gemspec** specifies: 
```ruby
s.add_dependency 'equivalent-xml', '~> 0.2.2'
```
which allows the version to slide up to `0.2.9` at the most current.  This is still about 3 years old and causes a blocker with any other component that wishes to use anything from the series of the 3 most recent major versions of `equivalent-xml`.  That is to say, `0.5.x` is current.  Among the features added are rspec 3 compatibility.  

We need to update.

atz:
Fixed in `5.0.0` if not before.

PR #211, opened 2016-03-31
use dor-services-app SDR proxy for getting preserved file content

jmartin-sul:
* add Sdr::Client.get_preserved_file_content:  have Dor::Contentable#get_preserved_file use that, instead of directly using content.sdr_server with basic auth in Dor::Contentable#get_preserved_file.
* tests for Dor::Contentable#get_preserved_file and Sdr::Client.get_preserved_file_content.

### DIFF
--- a/lib/dor/models/contentable.rb
+++ b/lib/dor/models/contentable.rb
@@ -62,12 +62,7 @@ module Dor
     end
 
     def get_preserved_file(file, version)
-      uri = URI(Config.content.sdr_server + "/sdr/objects/#{pid}/content/" + URI.encode(file) + "?version=#{version}")
-      req = Net::HTTP::Get.new(uri.request_uri)
-      req.basic_auth Config.content.sdr_user, Config.content.sdr_pass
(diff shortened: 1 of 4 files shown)


jmartin-sul:
was looking for lingering usage of `Config.content.sdr_user`, `Config.content.sdr_pass`, and `Config.content.sdr_server`.  found none, so will remove those from argo.  i think this is ready for review.

jmartin-sul:
if this were merged, i think it'd result in a minor version bump (i.e. 5.7.0 to 5.8.0, at the time of this comment).  it is a fix to existing behavior, but the `Dor::Contentable#get_preserved_file` now returns the content in a slightly different structure (the object returned by the proxy is immediately usable as a string, though it also has HTTP response methods available on it, see the new unit tests for usage).

jmartin-sul:
the endpoint used by this change was implemented in [dor-services app PR #25](https://github.com/sul-dlss/dor-services-app/pull/25)

ndushay:
:+1:

Issue #214, opened 2016-07-05
Indexing errors - Solr 6.1 rejecting workflow datestamps


LynnMcRae:
The indexing of workflow datestamps had a formatting error that seems to have been ignored to date, but which is now causing indexing failures. It was taking a time-adjusted datstamp like 2016-07-05T12:34:56.789-0700 and appending the value with a "Z". The resulting hybrid "...-0700Z" is not valid.

jmartin-sul:
fixed.  dor-services consumers are being upgraded.

if we backport this to 4.x, i'd be in favor of just reimplimenting by hand.  the change is straight-forward, and there was enough code cleanup and reformatting in the 4 to 5 upgrade that i'd be wary of conflicts if we were to cherry pick the commits from 5.x.

PR #215, opened 2016-07-06
workflow process status date solrization

jmartin-sul:
this fixes the way that dates for completed/errored workflow steps are added to the solr document, so that the dates are actually in the iso8601 format expected by solr (as opposed to an invalid format with a 'Z' appended to a date string that has a time zone offset).

also refines a bit of behavior to implement what was presumably intended (skipping non-existent dates).

tests added for both of those fixes.

deployed to stage, and was able to successfully index `druid:bb003xz2305`, which previously ran into the indexing error described in issue #214.

closes issue #214

### DIFF
--- a/lib/dor/workflow/document.rb
+++ b/lib/dor/workflow/document.rb
@@ -102,8 +102,8 @@ module Workflow
       processes.each do |process|
         next unless process.status.present?
         # add a record of the robot having operated on this item, so we can track robot activity
-        if process.date_time && process.status && (process.status == 'completed' || process.status == 'error')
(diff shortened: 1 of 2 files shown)


jmartin-sul:
argo-stage is using this code (by pointing to this branch of dor-services).  it appears to fix the problem.

could i get review from @ndushay and/or @atz if they have a moment?

ndushay:
LGTM. 👍

jmartin-sul:
in the interest of expediency, i'm going to take that thumbs up as license to merge this myself, so that we can release a new dor-services.

PR #217, opened 2016-07-07
add rights_descriptions_ssim to rights-related info in RightsMetadataDS#to_solr output, update tests

jmartin-sul:
* add rights_descriptions_ssim to RightsMetadataDS#to_solr output.  rights_descriptions_ssim aggregates rights info into an _ssim field for indexing/faceting as human readable strings.
* rights_metadata_spec.rb: update tests for updated dor-rights-auth behavior and updated RightsMetadataDS#to_solr
* dor-services.gemspec: minimum dor-rights-auth version is now 1.2.0, need new fields from Dor::RightsAuth#index_elements

this is the dor-services part of the work described in sul-dlss/dor-rights-auth#17

### DIFF
--- a/spec/dor/rights_metadata_spec.rb
+++ b/spec/dor/rights_metadata_spec.rb
@@ -50,19 +50,73 @@ describe Dor::RightsMetadataDS do
     before :each do
       allow(OpenURI).to receive(:open_uri).with('https://purl-test.stanford.edu/bb046xn0881.xml').and_return('<xml/>')
     end
+
     it 'should have correct primary' do
       expect(Dor.logger).to receive(:warn).with(/Cannot index druid:bb046xn0881\.descMetadata/)
       doc = @item.to_solr
+
       expect(doc).to match a_hash_including(
-        'rights_primary_ssi'  => 'world_qualified',
-        'metadata_source_ssi' => 'DOR',
-        'title_tesim'         => ['Indianapolis 500'],
-        'rights_characteristics_ssim' => include(
-          'world_discover', 'has_group_rights', 'has_rule', 'group|stanford', 'world_read', 'world|no-download', 'profile:group1|world1'
+        'rights_primary_ssi'          => 'world_qualified',
+        'rights_descriptions_ssim'    => include(
+          'world_qualified', 'location: reading_room (no-download)', 'stanford', 'world (no-download)', 'dark (file)'
         ),
-        'use_license_machine_ssi' => 'by-nc'
+        'metadata_source_ssi'         => 'DOR',
+        'title_tesim'                 => ['Indianapolis 500'],
+        'rights_characteristics_ssim' => include(
+          'world_discover', 'has_group_rights', 'has_rule', 'group|stanford', 'location', 'location_with_rule',
+          'world_read', 'world|no-download', 'profile:group1|location1|world1', 'none_read_file'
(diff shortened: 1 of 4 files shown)


spec/dor/rights_metadata_spec.rb
```diff
@@ -50,19 +50,73 @@
...
     end
+
     it 'should have correct primary' do
       expect(Dor.logger).to receive(:warn).with(/Cannot index druid:bb046xn0881\.descMetadata/)
       doc = @item.to_solr
+
       expect(doc).to match a_hash_including(
-        'rights_primary_ssi'  => 'world_qualified',
-        'metadata_source_ssi' => 'DOR',
-        'title_tesim'         => ['Indianapolis 500'],
```
jmartin-sul: `rights_primary_ssi`, `metadata_source_ssi`, `title_tesim` should be looking for the same field values only the formatting of the hash has changed w/ regards to these fields.

spec/dor/rights_metadata_spec.rb
```diff
@@ -50,19 +50,73 @@
...
+        'rights_primary_ssi'          => 'world_qualified',
+        'rights_descriptions_ssim'    => include(
+          'world_qualified', 'location: reading_room (no-download)', 'stanford', 'world (no-download)', 'dark (file)'
+        ),
+        'metadata_source_ssi'         => 'DOR',
+        'title_tesim'                 => ['Indianapolis 500'],
         'rights_characteristics_ssim' => include(
-          'world_discover', 'has_group_rights', 'has_rule', 'group|stanford', 'world_read', 'world|no-download', 'profile:group1|world1'
```
jmartin-sul: this has been updated to expect location related and dark file info

spec/dor/rights_metadata_spec.rb
```diff
@@ -50,19 +50,73 @@
...
+      expect(rights_md_ds).to receive(:dra_object).and_return(mock_dra_obj)
+
+      doc = rights_md_ds.to_solr
+      expect(doc).to match a_hash_including(
+        'obj_rights_locations_ssim'  => ['location'],
+        'file_rights_locations_ssim' => ['file_specific_location'],
+        'obj_rights_agents_ssim'     => ['agent'],
+        'file_rights_agents_ssim'    => ['file_specific_agent']
       )
-      expect(doc).not_to include('rights_errors_ssim')  # don't include empties
```
ndushay: is there an analogous don't include empties test that could/should be written?

ndushay:
I'm again a little out of my depth here with the specifics. Just have the one line comment with an overview perspective.

spec/dor/rights_metadata_spec.rb
jmartin-sul: well, that expectation was expanded to encompass the new fields also (see line 73 in the new code).  though there isn't a specific test for empties in general.  so... i think this keeps the spirit of the existing test (in that it tests the empty field behavior, though not as thoroughly as it could, since the old code could've also tested empty behavior for `rights_characteristics_ssim`).

i think this is adequate (no worse than it was before), but i'm happy to break that out into a separate (and more thorough) test if you think it's worth it.

ndushay: ok, now i see it.  we're good here.

PR #225, opened 2016-08-15
Adds publish_notify_on_success to touch change files in the PURL filesystem

### LINKED ISSUE #224: Publishable.publish_metadata needs to create a changes file on success
After a successful publish, the `publish_metadata` method needs to:
- create /purl/recent_changes/aa111bb2222 file via touch
- ensure that /purl/document_cache/.deletes/aa111bb2222 is removed

See https://github.com/sul-dlss/purl-fetcher/issues/103 for background.

drh-stanford:
This PR fixes #224.  It requires the application to add a new configuration parameter: `stacks.local_recent_changes`. This value is usually `/purl/recent_changes`.

@tingulfsen @jmartin-sul we need this PR reviewed for the Discovery Indexing sprint with Ben

### DIFF
--- a/spec/dor/publishable_spec.rb
+++ b/spec/dor/publishable_spec.rb
@@ -327,6 +327,7 @@ describe Dor::Publishable do
           expect(Dor::DigitalStacksService).to receive(:transfer_to_document_store).with('druid:ab123cd4567', /<oai_dc:dc/, 'dc')
           expect(Dor::DigitalStacksService).to receive(:transfer_to_document_store).with('druid:ab123cd4567', /<publicObject/, 'public')
           expect(Dor::DigitalStacksService).to receive(:transfer_to_document_store).with('druid:ab123cd4567', /<mods:mods/, 'mods')
+          expect(@item).to receive(:publish_notify_on_success).with(no_args)
         end
         it 'identityMetadta, contentMetadata, rightsMetadata, generated dublin core, and public xml' do
           @item.rightsMetadata.content = "<rightsMetadata><access type='discover'><machine><world/></machine></access></rightsMetadata>"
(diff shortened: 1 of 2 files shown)


spec/dor/publishable_spec.rb
```diff
@@ -340,6 +341,28 @@ class ItemizableItem < ActiveFedora::Base
...
+        @item.publish_notify_on_success
+      end
+      it 'writes empty file even when given only the base id' do
+        Dor::Config.push! {|config| config.stacks.local_recent_changes dir}
+        expect(File).to receive(:directory?).with(dir).and_return(true)
+        expect(@item).to receive(:pid).and_return('aa111bb2222')
+        expect(FileUtils).to receive(:touch).with(dir + '/aa111bb2222')
+        @item.publish_notify_on_success
+      end
+      it 'raises error if not configured' do
```
jmartin-sul: i might describe this as `misconfigured` or `not configured properly` instead of `not configured`, since it seems like this error could come up if the configuration were present but pointing to a non-existent/inaccessible dir.

jmartin-sul:
one very minor nitpick, otherwise, looks good to me...

do we know which of the dor-services consumers call `publish_metadata`?

drh-stanford:
Yes, `common-accessioning` and `dor-services-app` are the ones that use `publish_metadata` so they need updating when we cut a new dor-services release (5.10.0?).

Argo uses `publish_metadata_remotely` which calls `dor-services-app` via HTTP API.

I force pushed some naming cleanup on the specs based on your comments.

jmartin-sul:
cool, sounds good to me... the looking i did was very cursory, so `publish_metadata_remotely` was the only thing i'd come across, and i figured it was something like what you're describing above.

i'd be fine with the new version being 5.10.0.  this seems like a small change in intended behavior, so minor release seems appropriate.

PR #228, opened 2016-09-02
when publishing a purl, remove any existing .deletes record

peetucket:
fix for sul-dlss/purl-fetcher#90

### DIFF
--- a/lib/dor/models/publishable.rb
+++ b/lib/dor/models/publishable.rb
@@ -86,7 +86,11 @@ module Dor
       raise ArgumentError, "Missing local_recent_changes directory: #{local_recent_changes}" unless File.directory?(local_recent_changes)
       id = pid.gsub(/^druid:/, '')
       FileUtils.touch(File.join(local_recent_changes, id))
-      DruidTools::Druid.new(id, Dor::Config.stacks.local_document_cache_root).deletes_delete_record
+      begin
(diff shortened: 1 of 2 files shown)


peetucket:
Note: In addition to adding a new test, I am proposing to alter some existing related tests to actually look for the existence (or non-existence) of the relevant changes files instead of just expecting method calls to occur.

drh-stanford:
We may want to test what happens if `deletes_delete_record ` raises `Errno::EACCES` for permission denied. That's a somewhat plausible error case. That method doesn't use `FileUtils.rm_f` to force removal.

Otherwise, 👍

peetucket:
Interesting, do you have ideas for how to test that scenario?  I can't figure out how to set permissions for the .deletes file in such a way to perform this test (and still cleanup afterwards).

drh-stanford:
I would try mocking it out. e.g., `expect(...).to receive(:deletes_delete_record).and_raise(Errno::EACCES)`. would that work? If not, you can try `chmod`'ing the file to `444`, test removing it without the force (should fail), and then `chmod` it back to 644 to remove it. I'm not sure of the specific conditions under which `FileUtils.rm` fails but `FileUtils.rm_f` works.

peetucket:
ok, i was able to add a test mocking a failed delete (actually changing permissions didn't seem to work)...I also assume the correct behavior is to allow publish to succeed with a logged warning, rather than blowing up completely (and this is what the test confirms)

PR #235, opened 2016-09-27
remove extraneous displayType attribute from release tags

### LINKED ISSUE #226: Publish should not leave altered <release> tags in identityMetadata
A history of release activity is recorded in identityMetadata, a trail of true's and false's, when and by whom.  Publish determines the current release state of an item either from local release information or as inherited from a collection.  This answer is placed in a \<releaseData> container at the end of the Public XML.

Publish seem to also be replacing the \<release> tag in identityMetadata with the same or similar information. This should not be there ... the \<release> tag should simply not appear in identityMetadata in the Public XML. 

This was fixed at one time, by https://jirasul.stanford.edu/jira/browse/STNFRDSYNC-128, but still seems to be happening.  Test case = https://purl-test.stanford.edu/bb104kn2363.xml

### LINKED ISSUE #229: releasable adds two extraneous attributes to the release tag in identityMetadata
When you release an object, you get a statement like this;

```
<release displayType="file" release="true" to="SearchWorks" what="self" when="2016-09-16T19:28:05Z" who="pjreed">true</release>
```
Both the **displayType="file"** and **release="true"** attributes are incorrect, one deprecated, one just wrong.

peetucket:
and do not allow other extraneous attributes from being added to release tags

also, do not include release tag information in public XML identity metadata

fixes #229 and fixes #226 (proxy tickets: sul-dlss/discovery-indexing#31 and sul-dlss/discovery-indexing#26)

### DIFF
--- a/spec/dor/publishable_spec.rb
+++ b/spec/dor/publishable_spec.rb
@@ -91,8 +91,9 @@ describe Dor::Publishable do
         expect(@item).to receive(:released_for).and_return({})
         @p_xml = Nokogiri::XML(@item.public_xml)
       end
-      it 'does not include a releaseData element' do
+      it 'does not include a releaseData element and any info in identityMetadata' do
         expect(@p_xml.at_xpath('/publicObject/releaseData')).to be_nil
+        expect(@p_xml.at_xpath('/publicObject/identityMetadata/release')).to be_nil
       end
     end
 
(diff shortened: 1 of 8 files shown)


spec/dor/publishable_spec.rb
```diff
@@ -180,11 +181,12 @@ class ItemizableItem < ActiveFedora::Base
         expect(releases.map{ |r| r['to']}).to eq ['Searchworks', 'Some_special_place']
       end
 
-      it 'does include a releaseData element when there is content inside it' do
+      it 'include a releaseData element when there is content inside it, but does include this release data in identityMetadata' do
```
lmcglohon: Shouldn't this comment be "but does not include this release data in identityMetadata"?

peetucket: oops, yes, i will fix

PR #237, opened 2016-09-30
add a new thumb method to the object to parse the thumb image

### LINKED ISSUE #236: Publish - add <thumb> attribute to Purl Public XML
There is an enhancement to the rules to determine a thumbnail image for an object.  It is the introduction of a thumb=yes attribute that gives the object creator control over what image is used as a thumbnail, e.g., title page of the book instead of the first image which might be a generic leather cover or a blank flyleaf. For the full Rules of Thumb, see https://consul.stanford.edu/display/chimera/The+Rules+of+Thumb

In order to consolidate and make consistent the implementation of these now enhanced rules, we want to compute the thumbnail once during publish so that the answer is explicit in the Public XML to be picked up and used, no rules to implement. So publish should determine the thumbnail, if any, to be used for the object, and add the druid and that file name as a <thumb> attribute to the end of the Public XML, e.g., 
```
   <thumb>bk417pg0950/pc0065_b02_f07_i015.jp2</thumb>
```
Note that the druid may be different from the current object's druid in the case of an externalFile reference for a virtual resource.

Three primary use cases, processes that currently independently hunt for a thumb but do not implement the full rules: 

- Searchworks indexes the druid and file-name together for use in its UI, Revs as well?:
```
"file_id": [
   "bk417pg0950%2Fpc0065_b02_f07_i015.jp2"
],
```

- The Release processes inserts a thumb reference into the SDR managed 856 tag so that it can be indexed for catkey items in the same way (this requires moving this process from DOR to Purl, which resolves several issues):
```
856: u| http://purl.stanford.edu/hv987gp1617 x| SDR-PURL x| item x| file:bk417pg0950%2Fpc0065_b02_f07_i015.jp2 x| collection:xh235dd9059::David Rumsey Map Collection at Stanford University Libraries
```

- IIIF Manifests expresses it as a full stacks URL
```
"thumbnail": {
   "@id": "https://stacks.stanford.edu/image/iiif/bk417pg0950%2Fpc0065_b02_f07_i015/full/!400,400/0/default.jpg",
   "@type": "dcterms:Image",
   "format": "image/jpeg",
   "service": {
   "@context": "http://iiif.io/api/image/2/context.json",
   "@id": "https://stacks.stanford.edu/image/iiif/bk417pg0950%2Fpc0065_b02_f07_i015",
   "profile": "http://iiif.io/api/image/2/level1.json"
}
},
```
These will need to discard their current hunt for a thumb and simply pick up the value from the XML.

peetucket:
given set of rules specified here: https://consul.stanford.edu/display/chimera/The+Rules+of+Thumb

also add thumb value to public xml if it exists

fixes #236 and sul-dlss/discovery-indexing#11

### DIFF
--- a/lib/dor/models/publishable.rb
+++ b/lib/dor/models/publishable.rb
@@ -10,6 +10,58 @@ module Dor
     include Itemizable
     include Rightsable
 
+    # Compute the thumbnail for this object following the rules at https://consul.stanford.edu/display/chimera/The+Rules+of+Thumb
+    # @return [String] the computed thumb filename, with the druid prefix and a slash in front of it, e.g. oo000oo0001/filenamewith space.jp2
+    def thumb
+       return if contentMetadata.nil?
+       cm = contentMetadata.ng_xml
+       mime_type_finder = "@mimetype='image/jp2' or @mimeType='image/jp2'" # allow the mimetype attribute to be lower or camelcase when searching to make it more robust
+       thumb_image=nil
+
+       # these are the finders we will use to search for a thumb resource in contentMetadata, they will be searched in the order provided, stopping when one is reached
+       thumb_xpath_finders = [
+           {image_type: 'local', finder: "/contentMetadata/resource[@type='thumb' and @thumb='yes']/file[#{mime_type_finder}]"},      # first find a file of mimetype jp2 explicitly marked as a thumb in the resource type and with a thumb=yes attribute
+           {image_type: 'external', finder: "/contentMetadata/resource[@type='thumb' and @thumb='yes']/externalFile[#{mime_type_finder}]"}, # same thing for external files
+           {image_type: 'local', finder: "/contentMetadata/resource[(@type='page' or @type='image') and @thumb='yes']/file[#{mime_type_finder}]"},# next find any image or page resource types with the thumb=yes attribute of mimetype jp2
+           {image_type: 'external', finder: "/contentMetadata/resource[(@type='page' or @type='image') and @thumb='yes']/externalFile[#{mime_type_finder}]"},# same thing for external file
+           {image_type: 'local', finder: "/contentMetadata/resource[@type='thumb']/file[#{mime_type_finder}]"}, # next find a file of mimetype jp2 or resource type=thumb but no marked with the thumb directive
+           {image_type: 'external', finder: "/contentMetadata/resource[@type='thumb']/externalFile[#{mime_type_finder}]"}, # same thing for external file
+           {image_type: 'local', finder: "/contentMetadata/resource[@type='page' or @type='image']/file[#{mime_type_finder}]"}, # finally find the first very first page or image resource of mimetype jp2
+           {image_type: 'external', finder: "/contentMetadata/resource[@type='page' or @type='image']/externalFile[#{mime_type_finder}]"} # same thing for external file
+         ]
+              
+       thumb_xpath_finders.each do |search_path|
+         thumb_files = cm.xpath(search_path[:finder]) # look for a thumb  
+         if thumb_files.size > 0   # if we find one, return the filename based on whether it is a local file or external file
+           if search_path[:image_type] == 'local'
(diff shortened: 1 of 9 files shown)


lib/dor/models/publishable.rb
```diff
@@ -10,6 +10,58 @@ module Publishable
...
+       thumb_xpath_finders = [
+           {image_type: 'local', finder: "/contentMetadata/resource[@type='thumb' and @thumb='yes']/file[#{mime_type_finder}]"},      # first find a file of mimetype jp2 explicitly marked as a thumb in the resource type and with a thumb=yes attribute
+           {image_type: 'external', finder: "/contentMetadata/resource[@type='thumb' and @thumb='yes']/externalFile[#{mime_type_finder}]"}, # same thing for external files
+           {image_type: 'local', finder: "/contentMetadata/resource[(@type='page' or @type='image') and @thumb='yes']/file[#{mime_type_finder}]"},# next find any image or page resource types with the thumb=yes attribute of mimetype jp2
+           {image_type: 'external', finder: "/contentMetadata/resource[(@type='page' or @type='image') and @thumb='yes']/externalFile[#{mime_type_finder}]"},# same thing for external file
+           {image_type: 'local', finder: "/contentMetadata/resource[@type='thumb']/file[#{mime_type_finder}]"}, # next find a file of mimetype jp2 or resource type=thumb but no marked with the thumb directive
```
peetucket: yup

lib/dor/models/publishable.rb
```diff
@@ -10,6 +10,58 @@ module Publishable
...
+           {image_type: 'external', finder: "/contentMetadata/resource[@type='thumb']/externalFile[#{mime_type_finder}]"}, # same thing for external file
+           {image_type: 'local', finder: "/contentMetadata/resource[@type='page' or @type='image']/file[#{mime_type_finder}]"}, # finally find the first very first page or image resource of mimetype jp2
```
peetucket: first, i think you are firstly correct

lib/dor/models/publishable.rb
```diff
@@ -10,6 +10,58 @@ module Publishable
...
+           if search_path[:image_type] == 'local'
+             thumb_image="#{remove_druid_prefix}/#{thumb_files[0]['id']}"
+           else
+             thumb_image="#{remove_druid_prefix(thumb_files[0]['objectId'])}/#{thumb_files[0]['fileId']}"
+           end
+           break  # break out of the loop so we stop searching
+         end
+       end 
+              
+       thumb_image
```
peetucket: agreed that caching in memory might cause odd problems if contentMetadata changes, but using a local variable in the method so avoid multiple calls to thumb is totally reasonable and i'll do that

lib/dor/models/publishable.rb
```diff
@@ -10,6 +10,58 @@ module Publishable
...
+    end
+
+    # Return a URI encoded version of the thumb image for use by indexers (leaving the extension of the filename)
+    # @return [String] URI encoded version of the thumb with the druid prefix, e.g. oo000oo0001%2Ffilenamewith%20space.jp2
+    def encoded_thumb
+      return unless thumb
+      thumb_druid=thumb.split('/').first # the druid (before the first slash)
+      thumb_filename=thumb.split(/[a-zA-Z]{2}[0-9]{3}[a-zA-Z]{2}[0-9]{4}[\/]/).last # everything after the druid
```
peetucket: agreed it would be nice to centralize, and could be easy as a something like 

```def pid_regex 
  "[a-zA-Z]{2}[0-9]{3}[a-zA-Z]{2}[0-9]{4}"
end```

and then substitution that in as needed.

But where to put it?  Are there any modules that are available everywhere?  Should there be one?  Same concept applies to the "remove_druid_prefix" method...it is applicable broadly but it has to go somewhere and I want to be sure it is always available.  At first I put the thumb stuff in "contentable", but that isn't available to the Dor::Collection...which surprisingly, can have a thumb and can have contentMetadata...

PR #238, opened 2016-09-30
Add the public mods descriptive metadata to the publicObject document

cbeer:
This is expected to improve performance for applications and services that currently need to separately request the public xml and mods documents (exhibits, searchworks + revs indexing, purl itself, embed) with limited impact on applications that only need the public xml (stacks maybe the only consumer, although any performance degradation ought to be mitigated by better cache hit times).

We'll continuing publishing the stand-alone mods document, until such a time that everything has been republished (and the PURL code updated to use only the public xml document), or never (like the other stand-alone files we're producing that aren't consumed).


cbeer:
I've floated this by @LynnMcRae, @blalbrit, and @lmcglohon, who have their 👍  to the idea of doing this, at least.

PR #243, opened 2016-10-14
Update to support active-fedora 6 - 8

cbeer:
The major changes in AF 7+ that affect dor-services are:

- `mediashelf-loggable` (which monkey-patches all `Object`s with a `#logger` instance) is no longer included.
- AF8 adds a datastream prefix to indexed fields by default; while this is probably a good idea, I've disabled this behavior for now
- `ActiveFedora::Base.find` is now aware of class hierarchy, so you can't call `Dor::Item.find` on e.g. a `Dor::Collection` and have it work right. 
- Upstream `#to_solr` no longer mutates its arguments.

Fortunately, the changes seem to be backwards compatible with AF 6.x. I don't know whether this should indicate a new major release or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,26 @@ notifications:
   email: false
 
 rvm:
-  - 2.1.2
-  - 2.2.2
+  - 2.2.5
   - 2.3.1
+
+env:
+  - 'RAILS_VERSION="~> 4.2"'
+  - 'RAILS_VERSION="~> 5.0"'
+
+matrix:
+  include:
+    - env: 'RAILS_VERSION="~> 4.2"'
+      rvm: 2.1.2
+    - env:
+        - 'AF_VERSION="~> 6.0"'
+        - 'RAILS_VERSION="~> 4.2"'
+      rvm: 2.3.1
+    - env:
+        - 'AF_VERSION="~> 7.0"'
+        - 'RAILS_VERSION="~> 4.2"'
+      rvm: 2.3.1
+    - env:
+        - 'AF_VERSION="~> 8.0"'
+        - 'RAILS_VERSION="~> 4.2"'
+      rvm: 2.3.1
--- a/Gemfile
+++ b/Gemfile
@@ -6,3 +6,17 @@ end
 
 # Dependencies are defined in dor-services.gemspec
 gemspec
+
+gem 'activemodel', ENV['RAILS_VERSION'] if ENV['RAILS_VERSION']
+gem 'active-fedora', ENV['AF_VERSION'] if ENV['AF_VERSION']
+
+
+# Due to a possible bundler bug, resolving the dependencies for linkeddata 1.99
+# either breaks or takes an incredibly long time. Pinning these dependencies
+# seem to make it work.
+gem 'linkeddata', '~> 1.99'
+gem 'rdf', '~> 1.99'
+gem 'ebnf', '1.0.0'
+gem 'rdf-reasoner', '~> 0.3.0'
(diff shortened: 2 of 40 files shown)


mejackreed:
When looking through this, things are looking good. I'm 👍  for merging, though its hard to answer the semver question without knowing the context of where this sits. @atz ?

jmartin-sul:
i'd like to test this with argo (and maybe dor-indexing-app), esp if we're not going to bump dor-services to v6 (i'm not really sure either way at this point whether that seems like the right thing to do to me).

for instance:
> `ActiveFedora::Base.find` is now aware of class hierarchy, so you can't call `Dor::Item.find` on e.g. a `Dor::Collection` and have it work right.

argo's definitely been guilty of this in the past, and while i can't think of a specific instance off-hand where it's still happening, i'd be a bit surprised if it's not at all anymore (ditto other dor-services consumers).  that's not to say that should hold us up from this upgrade -- that's definitely bad behavior on argo's part, and should get fixed.  but it'd be good to be aware of things like that (and it shouldn't be too hard to grep for specific object type loads).  maybe we could deploy to stage and do a bulk reindexing of it's 80k or so objects, and see what errors we come up with?

after writing the above, i feel not-too-strongly like this should result in dor-services v6.  though since we're currently supporting v4 and v5, that would mean either supporting 3 major versions, or dropping support for version 4.  i don't think this is as big of a breaking change as v4 to v5, but that was a pretty massive overhaul, so that's not the best standard to use.

in addition to @atz, does @drh-stanford have opinions?  who else has dealt lots with dor-services usage?  @peetucket?

cbeer:
Note that I already have a PR that fixes Argo (at least the bits covered by tests): https://github.com/sul-dlss/argo/pull/844.

peetucket:
so the core issue is that consumers need to be sure of the object type they are loading and instantiate the correct object type?  what if they don't know ahead of time but just have a druid?  is there a generic object you can load?

cbeer:
Yes, `Dor.find` (as I switched all these to) or `ActiveFedora::Base.find` both work just fine. The problem is when you use a specific class to try to load something that isn't that type of object (e.g. using `Dor::Item.find` to try to load a `Dor::Collection`).

peetucket:
Ok, that's good to know. So that's a pretty easy upgrade path for someone who gets bitten.  Let's put that in the release notes.

cbeer:
`ENABLE_SOLR_UPDATES` is still a thing (even in ActiveFedora master). We just don't need to test that it works in dor-services, as it is tested upstream just fine.

jmartin-sul:
i didn't realize (until seeing it in slack this morning) that this is still compatible with the older version of ActiveFedora. since there's still backwards compatibility, and since the upgrade path seems reasonably clear and not all that burdensome, i'm 👍 on this (having what looks like a good sweep of the affected behavior in argo is a big help too).

and the behavior that's not backwards compatible for consumers that upgrade ActiveFedora is a thing that should get fixed anyway, so it doesn't seem terrible that this provides pressure to make that change.

i'd vote for keeping this change on the 5.x line, since there's backwards compatibility.  if the consumer just needs to pin something in `Gemfile` to keep from breaking on upgrade, it seems reasonable for that to fall under a minor version bump.

if it comes to that, i would also vote for trying to officially desupport v4 before we started supporting a v6.  and i'd be in favor of bumping to v6 once AF6 is no longer supported by dor-services, since that does seem like a major breaking change (albiet not so much as a wholesale overhaul of what's indexed to what field names).

PR #244, opened 2016-10-14
skip checking release tags on collections that an item is in if it is…

peetucket:
… the same as the item itself; unskip some tests and remove one that was duplicative

fixes sul-dlss/discovery-indexing#14

### DIFF
--- a/lib/dor/models/releaseable.rb
+++ b/lib/dor/models/releaseable.rb
@@ -79,7 +79,8 @@ module Dor
     def get_release_tags_for_item_and_all_governing_sets
       return_tags = release_nodes || {}
       collections.each do |collection|
-        return_tags = combine_two_release_tag_hashes(return_tags, Dor.find(collection.id).get_release_tags_for_item_and_all_governing_sets) # recurvise so parents of parents are found
+        next if collection.id == id  # recursive, so parents of parents are found, but we need to avoid an infinite loop if the collection references itself (i.e. bad data)
+        return_tags = combine_two_release_tag_hashes(return_tags, Dor.find(collection.id).get_release_tags_for_item_and_all_governing_sets)
       end
       return_tags
     end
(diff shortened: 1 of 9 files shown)


lib/dor/models/releaseable.rb
```diff
@@ -79,7 +79,7 @@ def get_self_release_tags(tags)
     def get_release_tags_for_item_and_all_governing_sets
       return_tags = release_nodes || {}
       collections.each do |collection|
-        return_tags = combine_two_release_tag_hashes(return_tags, Dor::Item.find(collection.id).get_release_tags_for_item_and_all_governing_sets) # recurvise so parents of parents are found
+        return_tags = combine_two_release_tag_hashes(return_tags, Dor::Item.find(collection.id).get_release_tags_for_item_and_all_governing_sets) unless collection.id == id # recurvise so parents of parents are found, but avoid an infinite loop if the collection references itself
```
atz: Probably easier to read with a guard clause.  I.E.: `next if collection.id == id`.  Also, the comment has `recurvise`.

atz: Ideally these release tag hashes would have a thin class of their own, a descendant of Hash overriding `merge`, and then all the semantics here would be pretty clear.

peetucket: agreed on point # 1, fixed

point # 2 is also good but that refactor is probably outside the time scope we want to work with right now

mejackreed: point #2.. ran into this as well. Is it a by design thing in DorServices to not use any classes besides the main classes? `Dor::Item`, etc.. 

I found it really difficult to work with this module because everything returned a special data structure hash or array thingy

mejackreed: Agreed.. didn't want to muddy the waters in this PR.. just a comment that I ran into similar things.. Thanks @peetucket

PR #245, opened 2016-10-26
enhanced rights setting

jmartin-sul:
enhanced setting of object rights and APO-specified default object rights:
* allow location based rights
* allow specification of the no-download rule

~~marking WIP because i'd like to add a couple new tests for the newly settable rights types.~~

### DIFF
--- a/lib/dor/datastreams/rights_metadata_ds.rb
+++ b/lib/dor/datastreams/rights_metadata_ds.rb
@@ -82,7 +82,6 @@ module Dor
       end
 
       label = rights == 'dark' ? 'none' : 'world'
-      @dra_object = nil # until TODO complete, we'll expect to have to reparse after modification
       rights_xml.search('//rightsMetadata/access[@type=\'discover\']/machine').each do |node|
         node.children.remove
         node.add_child Nokogiri::XML::Node.new(label, rights_xml)
@@ -110,6 +109,7 @@ module Dor
         end
       end
 
+      @dra_object = nil # until TODO complete, we'll expect to have to reparse after modification
       self.content = rights_xml.to_xml
       content_will_change!
     end
--- a/lib/dor/datastreams/rights_metadata_ds.rb
+++ b/lib/dor/datastreams/rights_metadata_ds.rb
@@ -83,7 +83,40 @@ module Dor
       RightsMetadataDS.valid_rights_types.include? rights
     end
 
-    # @param rights [string] archetypical rights to assign: 'world', 'stanford', 'none' or 'dark'
+    # a helper method for setting up well-structured rights_xml based on a rights type code
+    # @param rights_xml [ng_xml] a nokogiri xml (ruby) object that represents the rights xml for a DOR object
+    # @param rights_type [string] a recognized rights type code ('world', 'dark', 'loc:spec', etc)
+    def self.upd_rights_xml_for_rights_type(rights_xml, rights_type)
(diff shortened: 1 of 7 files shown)


lib/dor/datastreams/rights_metadata_ds.rb
```diff
@@ -61,36 +61,77 @@ def dra_object
...
+        'world' => 'World',
+        'world-nd' => 'World (no-download)',
+        'stanford' => 'Stanford',
+        'stanford-nd' => 'Stanford (no-download)',
+        'loc:spec' => 'Location spec',
+        'loc:music' => 'Location music',
+        'dark' => 'Dark (Preserve Only)',
+        'none' => 'Citation Only'
+      }
+    end
```
ndushay: This could actually be a class constant, no?

jmartin-sul:
ok, i've added a bunch of tests, and i believe all the concerns have been addressed:
* thorough method level tests for various permutations of `set_rights_xml_for_rights_type`
* tests in editable_spec.rb indicating "how `default_rights_for_indexing` differs from `default_rights` and that `to_solr` uses the former and not the latter"
* settability of no-download and location-based rules in registration_services_spec.rb
* turned the rights type code mapping for human readable strings into class constant

re-review from @ndushay?

may also be of interest to @atz and @tingulfsen

ndushay:
LGTM;  👍  for merging.

lib/dor/models/editable.rb
```diff
@@ -232,16 +232,51 @@ def use_license=(use_license_machine)
...
+          'world'
+        end
+      elsif machine_read_access_node && machine_read_access_node.search('./location[text()="spec"]').length == 1
+        'loc:spec'
+      elsif machine_read_access_node && machine_read_access_node.search('./location[text()="music"]').length == 1
+        'loc:music'
+      elsif machine_discover_access_node && machine_discover_access_node.search('./world').length == 1
+        'none'
+      elsif machine_discover_access_node && machine_discover_access_node.search('./none').length == 1
+        'dark'
```
tingulfsen: So, this reads a little funny to me - can you explain what's happening in lines 257-260, @jmartin-sul ?

tingulfsen:
I like it, but would like the question regarding lib/dor/models/editable.rb to be explained.

lib/dor/models/editable.rb
jmartin-sul: i was just trying to cut down on the full xpath searches and samey xpath statements (particularly, what would've been some long and similar xpath queries for the `stanford` rights types).  also wanted to make more explicit that the conditions are working within either the read access or discover access nodes.  and since it's possible that either the read access node or the discover access node might be `nil`, i was guarding against referencing a method on a `nil` object with the `machine_read_access_node &&` and `machine_discover_access_node &&` parts of the conditions.

jmartin-sul: ah, sorry, specifically in 257-260:  if none of the prior three top level branches are true (it's not a stanford restricted object, it's not a world readable object, it's not a location restricted object), then it must be either "citation only" or "dark".  if it's world discoverable, it's citation only.  if it's not discoverable, it's dark.  if the XML doesn't fit any of those patterns, it's in a structure we don't support/understand, and we can't return a corresponding rights type code.

jmartin-sul: fixed up the comments there and pushed the slightly updated version

PR #246, opened 2016-11-04
add indexing benchmark metrics for load_instance and to_solr calls

drh-stanford:
This PR is connected to https://github.com/sul-dlss/dor_indexing_app/issues/49 and it implements some simple benchmark timings for the indexer's `Dor.load_instance` and `obj.to_solr` calls using the Ruby built-in `benchmark` API.

The log output looks like this in `indexing.log`:

```
I, [2016-11-08T11:48:51.413245 #8790]  INFO -- : successfully updated index for druid:aa111bb2222 (metrics: load_instance realtime 0.000022s total CPU 0.000000s; to_solr realtime 0.000039s total CPU 0.000000s)
```

Note that I did not move the location of the `logger.info` command but it logs *before* it actually adds the document to Solr via `RSolr.add`.

### DIFF
--- a/lib/dor/services/indexing_service.rb
+++ b/lib/dor/services/indexing_service.rb
@@ -1,3 +1,5 @@
+require 'benchmark'
+
 module Dor
   class IndexingService
     ##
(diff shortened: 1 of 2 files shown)


lib/dor/services/indexing_service.rb
```diff
@@ -55,9 +57,11 @@ def self.reindex_pid(pid, *args)
...
       end
 
-      obj = Dor.load_instance pid
-      solr_doc = reindex_object obj, options
-      index_logger.info "updated index for #{pid}"
+      obj = nil
+      load_stats = Benchmark.measure('load_instance') { obj = Dor.load_instance pid }
+      solr_doc = nil
+      to_solr_stats = Benchmark.measure('to_solr') { solr_doc = reindex_object obj, options }
+      index_logger.info "successfully updated index for #{pid} (metrics: #{load_stats.format('%n %rs %ts')}; #{to_solr_stats.format('%n %rs %ts')}; total #{load_stats.total + to_solr_stats.total}s)"
```
jmartin-sul: i think it'd be nice to include a short comment inline about what's being written out in the `format` statement, and what the `total` indicates (i assume total runtime for the benchmarked operation?).

drh-stanford:
Thanks for the comments. I cleaned up the code and repushed. I changed the format to make it clear what the timings mean in the log itself. I revised the PR comment to that effect.

PR #247, opened 2016-11-07
use new_record? instead of new_object?

mejackreed:
Addresses some of the versionable issues seen in Argo

 - https://github.com/sul-dlss/argo/issues/863
 - https://github.com/sul-dlss/argo/issues/864


cbeer:
RSpec's verifying doubles should be able to catch errors like this. I remember looking at it (perhaps in the context of dor-services) and didn't get great results given the way we've set up our tests, though. Maybe it's worth a try?

PR #248, opened 2016-11-10
Add Dor::Identifiable#adapt_to_cmodel

cbeer:
overriding ActiveFedora's method, to inspect identityMetadata for the object class to use

Fixes https://github.com/sul-dlss/dor_indexing_app/issues/82

### DIFF
--- a/lib/dor/models/identifiable.rb
+++ b/lib/dor/models/identifiable.rb
@@ -224,6 +224,19 @@ module Dor
       result.nil? ? druid : result[0]  # if no matches, return the string passed in, otherwise return the match 
     end
 
+    # Override ActiveFedora::Core#adapt_to_cmodel (used with associations, among other places) to
+    # preferentially use the objectType asserted in the identityMetadata.
+    def adapt_to_cmodel
+      object_type = identityMetadata.objectType.first
+      object_class = Dor.registered_classes[object_type]
+
+      if object_class
+        self.instance_of?(object_class) ? self : self.adapt_to(object_class)
+      else
+        super
+      end
+    end
+
     private
 
     def solrize_related_obj_titles(solr_doc, relationships, title_hash, union_field_name, nonhydrus_field_name, hydrus_field_name)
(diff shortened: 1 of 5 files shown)


lib/dor/models/identifiable.rb
```diff
@@ -224,6 +224,19 @@ def remove_druid_prefix(druid=id)
       result.nil? ? druid : result[0]  # if no matches, return the string passed in, otherwise return the match 
     end
 
+    # Override ActiveFedora::Core#adapt_to_cmodel (used with associations, among other places) to
+    # preferentially use the objectType asserted in the identityMetadata.
+    def adapt_to_cmodel
+      object_type = identityMetadata.objectType.first
+      object_class = Dor.registered_classes[object_type]
+
+      if object_class
```
drh-stanford: for my clarification, doesn't the `registered_classes[object_type]` call throw `NameError` if the `object_type` is missing? if so, I don't think this `if` condition is ever false. that is, when would `super` ever get called?

spec/dor/identifiable_spec.rb
```diff
@@ -312,4 +312,14 @@ class IdentifiableItem < ActiveFedora::Base
       expect(item.get_related_obj_display_title(mock_apo_obj, mock_default_title)).to eq(mock_default_title)
     end
   end
+
+  describe '#adapt_to_cmodel' do
+    context 'for a Hydrus collection' do
```
drh-stanford: i think it makes sense to add tests for an item adaption for clarity. item_druid_bb004bn8654.xml is a fixture for a Hydrus item and item_druid_dc235vd9662.xml is a fixture for a Dor item.

lib/dor/models/identifiable.rb
cbeer: I think it's just `nil`. We were guarding against it before too: https://github.com/sul-dlss/dor-services/pull/248/files#diff-0d90e7b0ff1134056873ce5aa4b3e0f1L26

spec/dor/identifiable_spec.rb
cbeer: 👌 . I don't think we need to be too exhaustive, although maybe I'm too optimistic about the coverage on `Dor.find`.

PR #250, opened 2016-11-11
Refactor DOR models to use community-accepted practices around class inheritence

cbeer:
All so we can use e.g. `Dor::Base.find` to retrieve any DOR model object, or `Dor::Set` to retrieve sets or collections..

### DIFF
--- a/lib/dor-services.rb
+++ b/lib/dor-services.rb
@@ -21,7 +21,7 @@ module Dor
     # index is missing the objectType property.
     # @param [String] pid The object's PID
     def load_instance(pid)
-      Dor::Base.find pid, cast: true
+      Dor::Abstract.find pid, cast: true
     end
 
     # Get objectType information from solr and load the correct class the first time,
@@ -121,7 +121,6 @@ module Dor
     # ActiveFedora Classes
     autoload :Abstract,          'dor/models/abstract'
     autoload :Agreement,         'dor/models/agreement'
-    autoload :Base,              'dor/models/base'
     autoload :Item,              'dor/models/item'
     autoload :Set,               'dor/models/set'
     autoload :Collection,        'dor/models/collection'
--- a/lib/dor-services.rb
+++ b/lib/dor-services.rb
@@ -135,6 +135,7 @@ module Dor
     # ActiveFedora Classes
     autoload :Abstract,          'dor/models/abstract'
     autoload :Agreement,         'dor/models/agreement'
+    autoload :Base,              'dor/models/base'
     autoload :Item,              'dor/models/item'
     autoload :Set,               'dor/models/set'
     autoload :Collection,        'dor/models/collection'
--- a/lib/dor-services.rb
+++ b/lib/dor-services.rb
@@ -21,9 +21,7 @@ module Dor
     # index is missing the objectType property.
     # @param [String] pid The object's PID
     def load_instance(pid)
-      obj = Dor::Abstract.find pid, cast: false
-      return nil if obj.new_record?
-      obj.adapt_to_cmodel
+      Dor::Base.find pid, cast: true
     end
 
     # Get objectType information from solr and load the correct class the first time,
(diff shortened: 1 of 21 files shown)


cbeer:
I've also extracted the first 3 commits, which I believe to be uncontroversial, into https://github.com/sul-dlss/dor-services/pull/251

lib/dor-services.rb
```diff
@@ -31,29 +29,15 @@ def load_instance(pid)
...
     # TODO: return enumerable and lazy load_instance
     # TODO: restrict fieldlist (fl) for non-:lightweight queries
     def find_all(query, opts = {})
       resp = SearchService.query query, opts
       resp.docs.collect do |solr_doc|
-        doc_version = solr_doc[INDEX_VERSION_FIELD].first rescue '0.0.0'
-        doc_version = Gem::Version.new(doc_version)
-        object_type = Array(solr_doc[ActiveFedora::SolrService.solr_name('objectType', :symbol)]).first
-        object_class = registered_classes[object_type] || ActiveFedora::Base
-        if opts[:lightweight] && doc_version >= Gem::Version.new('3.1.0')
```
cbeer: I haven't found anywhere (yet) that uses `:lightweight` in any useful way, especially with the fixes now in AF/Rubydora.

drh-stanford: should we deprecate `find_all`? the only place I see it being used now is to find workflow objects by name.

drh-stanford:
Might be helpful to put the class hierarchy in a comment near `Dor.find` or elsewhere documented. My understanding is this:

(Agreements) < Dor::Item < Dor::Abstract < Dor::Base < ActiveFedora::Base
(Items, Sets, Collections, APOs) < Dor::Abstract < Dor::Base < ActiveFedora::Base 
(WorkflowObjects) < Dor::Base < ActiveFedora::Base

lib/dor-services.rb
```diff
@@ -31,29 +29,15 @@ def load_instance(pid)
     # @param [String] pid The object's PID
     # @return [Object] the ActiveFedora-modeled object
     def find(pid, opts = {})
-      opts[:rows] = 1 # we know we are going to just pay attention to the .first doc anyway
-      find_all(%(id:"#{pid}"), opts).first || load_instance(pid)
+      load_instance(pid)
```
cbeer: Yes, it's one of them... and not necessarily a bad idea, just seems unnecessary if we immediately turn around and make other requests.

drh-stanford:
There was a conflict in the `Dor.find_all` method that I resolved. It now reads:

```ruby
    def find_all(query, opts = {})
      resp = SearchService.query query, opts
      resp['response']['docs'].collect do |solr_doc|
        find solr_doc['id']
      end
    end
```

cbeer:
@drh-stanford 👌 , looks good to me.

lib/dor-services.rb
cbeer: Sure, I'd be in favor of that. Using AF's native finders probably makes more sense.

cbeer:
screen shot 2016-11-14 at 5 43 00 pm

LynnMcRae:
I am curious why workflow objects don't come from Dor::Abstract like APOs. Is it the nature of the beast (a systems/operational meta-thingy)? or something specific about how they are constructed. Other than the addition of the workflowDefinition datastream, in theory it should act like a regular DOR object, with identityMetadata, descMetadata, preservable but not publishable or releasable. If it is structural irregularities (I see cases missing descMetadata for instance), I would rather make them conform. But if they belong separate for other reasons, then I'm good.

drh-stanford:
It is curious. `Dor::Abstract` just adds `Identifiable` which `WorkflowObject` uses. So I don't really see a reason why it can't base off of `Dor::Abstract`. It's likely to be historical as the last time that class inheritance was modified was by mbklein in 2012.

cbeer:
I don't know what the intent was, I was merely trying to accurately represent the current state of affairs without adding or removing any features. Currently, `Dor::WorkflowObject` is only `Identifiable` and `Governable` (and the modules that implies). `Dor::Abstract` adds `Describable`, `Processable`, and `Versionable`, and it isn't clear to me if those are desirable additions to `Dor::WorkflowObject`.

drh-stanford:
I see so in the refactor you surfaced that Abstract really pulls in Describable, etc. So we have a Base and an Abstract in the current codebase. I think keeping the refactor strictly a refactor is best at this point and the clearer class diagram and code hierarchy will allow us to restructure with more confidence later.

cbeer:
Exactly. I took common code and pushed it as far up the hierarchy as possible, without altering behavior. It's certainly somewhat attractive to avoid introducing `Dor::Base`, which may turn out to be only temporary, but unless it's obvious to @LynnMcRae that `Dor::WorkflowObject` should inherit all the common behaviors, I don't see any real harm in deferring that either.

LynnMcRae:
My offer to update workflow objects so they conform to Dor::Abstract stands, though that doesn't sound like it is needed. I would have no objection to workflow objects inheriting new behaviors to simplify the hierarchy but it is not imperative right now, so the status quo is also fine.  Note that 10 WF objects declare Dor::Item as a model in addition to Dor:WorkflowObject, for no good reason than object hand-copying I'm sure.

cbeer:
Ok, I've gone ahead and made `Dor::WorkflowObject` inherit from `Dor::Abstract`.

PR #252, opened 2016-11-11
Default to loading unknown objects as a Dor::Item instance

cbeer:
Fixes https://github.com/sul-dlss/dor_indexing_app/issues/93

### DIFF
--- a/lib/dor/models/identifiable.rb
+++ b/lib/dor/models/identifiable.rb
@@ -233,7 +233,20 @@ module Dor
       if object_class
         self.instance_of?(object_class) ? self : self.adapt_to(object_class)
       else
-        super
+        if ActiveFedora::VERSION < '8'
+          result = super
+          if result.class == Dor::Abstract
+            self.adapt_to(Dor::Item)
+          else
+            result
+          end
+        else
+          begin
+            super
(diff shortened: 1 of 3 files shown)


drh-stanford:
Looks like `ActiveFedora::ModelNotAsserted` is undefined in AF 6.x and 7.x. Travis builds failed...

cbeer:
Gah, just can't win... I'll just do it the hard way, then.

lib/dor/models/identifiable.rb
```diff
@@ -233,7 +233,20 @@ def adapt_to_cmodel
       if object_class
         self.instance_of?(object_class) ? self : self.adapt_to(object_class)
       else
```
drh-stanford: what if `identityMetadata` or `identityMetadata.objectType` are `nil`? (i.e., identityMetadata datastream is missing or lacking an objectType element. won't that throw a `NoMethodError` and cause `adapt_to_cmodel` to raise out rather than executing this block here.

cbeer: The `foxml_empty` test I added has no identityMetadata and falls through to the default as expected.

PR #254, opened 2016-11-11
Remove indirection to support multiple formats of descMetadata

cbeer:
MODS is now the only format we have, and many things would need to be reworked to support other formats. This removes some indirection from describable, hopefully enhancing the clarity of what's there.


drh-stanford:
I'm curious as to whether we're encoding the `mods` metadata format in `descMetadata` or `identityMetadata`. I'm ok with the PR but I'd like to know whether we're ignoing data (and assuming everything is MODS) and add a comment to that effect if so.

PR #256, opened 2016-11-11
Refactor to avoid unnecessarily retrieving collection information

cbeer:


cbeer:
Hopefully https://github.com/sul-dlss/dor_indexing_app/issues/97 will obviate the desire for this, but in the meantime, this preserves compatibility without making unnecessary requests.

PR #257, opened 2016-11-12
Refactor Preservable#build_provenanceMetadata_datastream

cbeer:
The `content=` is unnecessary; see  https://github.com/projecthydra/active_fedora/blob/8.x-stable/lib/active_fedora/datastreams/nokogiri_datastreams.rb#L50

### DIFF
--- a/lib/dor/models/preservable.rb
+++ b/lib/dor/models/preservable.rb
@@ -11,11 +11,9 @@ module Dor
 
     def build_provenanceMetadata_datastream(workflow_id, event_text)
       workflow_provenance = create_workflow_provenance(workflow_id, event_text)
-      dsname = 'provenanceMetadata'
-      ds = datastreams[dsname]
-      ds.label = 'Provenance Metadata' unless datastreams.keys.include?(dsname)
+      ds = datastreams['provenanceMetadata']
+      ds.label ||= 'Provenance Metadata'
       ds.ng_xml = workflow_provenance
-      ds.content = ds.ng_xml.to_s
       ds.save
     end
 


lib/dor/models/preservable.rb
```diff
@@ -11,11 +11,9 @@ module Preservable
 
     def build_provenanceMetadata_datastream(workflow_id, event_text)
       workflow_provenance = create_workflow_provenance(workflow_id, event_text)
-      dsname = 'provenanceMetadata'
-      ds = datastreams[dsname]
-      ds.label = 'Provenance Metadata' unless datastreams.keys.include?(dsname)
+      ds = datastreams['provenanceMetadata']
+      ds.label ||= 'Provenance Metadata'
       ds.ng_xml = workflow_provenance
-      ds.content = ds.ng_xml.to_s
```
jmartin-sul: i actually think this is still necessary for the content in the datastream to get updated properly.  there are lots of examples where we still do this, and IIRC, it's something that @atz went over pretty thoroughly and attempted to make consistent/correct in the dor-services 4 to 5 upgrade.

i'm also just looking at this chunk of code here, so maybe i don't have the full context and the content gets set elsewhere.  or maybe i'm misremembering and we actually need to get rid of this usage.  but there are lots of example that turn up if you do `git grep "content.*=.*to_s"`.

@atz?

cbeer: Ok. I'll take a look through all of those. I really hope they'll turn out to be unnecessary.

cbeer: I've pushed up https://github.com/sul-dlss/dor-services/pull/258, which should deal with all of those.. Next week, it's probably worth doing some testing in argo and other places to see if we can find (and then properly fix..) any issues.

cbeer:
Replaced by #258. No reason to just handle this one case, I think.

PR #258, opened 2016-11-12
Stop excessive `#content=` usage

cbeer:


cbeer:
@jmartin-sul I've tested this branch against Argo without problems. Anything else you think should be especially tested against it?

drh-stanford:
So, if I understand this change correctly the `ng_xml_will_change!` method will automatically update the `content` whenever a `save` happens? I poked around ActiveFedora 8.x but didn't see this method defined...

cbeer:
Yes, `ng_xml_will_change!` marks the datastream as dirty, updating the content.. I believe the logic is somewhere in `ActiveModel`?

drh-stanford:
Ok, I see it in `ActiveModel::Dirty`. It appears that this will save us redundant XML serialization. Am I correct about that?

cbeer:
Yes, that's the immediate improvement. This also helps us in following upstream conventions.

jmartin-sul:
@cbeer: apologies for the slow reply, but that all sounds good to me.  the code you pointed to in PR #257 makes sense, and if you did some manual testing against argo and datastream editing still looked fine, that's enough for me.  IIRC, it was pretty immediately apparent that the save wasn't being made when this was a problem in the past.

PR #260, opened 2016-11-14
Remove Dor::Upgradable

cbeer:
We haven't used the `Dor::Upgradable` framework in over 3 years. Is it possible it's been replaced by some other approach, or been rendered unnecessary?


drh-stanford:
It looks to me like it was for a one-time data migration process.

cbeer:
Right -- several years ago, it was used to remediate some data errors. Given we haven't used it since then, I'm not sure it's valuable enough to keep around.

PR #261, opened 2016-11-14
Remove unused descMetadata setters

cbeer:
I don't see where these setters are being used any more. All the downstream code I've found uses e.g. `Editable#mods_title=` instead.


drh-stanford:
I couldn't find any usage of these either.

PR #263, opened 2016-11-14
[WIP] Use ActiveFedora to manage constituent object relationships

### LINKED ISSUE #206: Dor::Item access to de-referenced constituents?
It would be really useful to be able to have access to a de-referenced object from the `isConstituentOf` RELS-EXT relationship.  It wasn't immediately obvious if this was available via `Dor::Item` or elsewhere already.

The ItemRelease workflow currently has some [code](https://github.com/sul-dlss/item-release/blob/3bb371f1649efb200b1d3e6ab8d7601ff21d24aa/lib/dor/update_marc_record_service.rb#L187-L191) to do this as a stop-gap, but it seems like this is something that would live in dor-services land (unless I'm way off-base).

cbeer:
Fixes #206

- [ ] Update ActiveFedora to avoid unnecessary calls to Solr, or
- [ ] Stub `ActiveFedora::SolrService` calls for retrieving `has_and_belongs_to_many` relations

### DIFF
--- a/lib/dor/models/contentable.rb
+++ b/lib/dor/models/contentable.rb
@@ -2,6 +2,10 @@ module Dor
   module Contentable
     extend ActiveSupport::Concern
 
+    included do
+      has_and_belongs_to_many :constituent_objects, property: :is_constituent_of, class_name: 'ActiveFedora::Base'
+    end
+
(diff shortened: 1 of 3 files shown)


atz:
@cbeer care to rebase this?

PR #264, opened 2016-11-14
Add additional indexed fields from MODS for increased parity with existing public_dc indexing

cbeer:


cbeer:
I suspect any such test would be pretty brittle, and is mostly testing if stanford-mods works (which we oughtn't need to do here..). Happy to be persuaded, but I'm not sure "nice" is emphatic enough.

lib/dor/datastreams/desc_metadata_ds.rb
```diff
@@ -29,6 +29,7 @@ class DescMetadataDS < ActiveFedora::OmDatastream
       t.scale       :index_as => [:symbol]
       t.topic       :index_as => [:symbol, :stored_searchable]
       t.abstract    :index_as => [:stored_searchable]
+      t.identifier  :index_as => [:symbol, :stored_searchable]
```
drh-stanford: is this used anywhere? or is it just for completeness?

cbeer: This is intended to eventually replace the `public_dc_identifier_tesim` field.

jmartin-sul:
@cbeer:  yeah, i do agree that such tests might be brittle... i'm not sure that having expectations about what stanford-mods returns would be a bad thing, but i can also see how implementing/maintaining that could be more of a headache than it's worth.

anyway, merging...

PR #265, opened 2016-11-15
Exclusively use MODS-derived display titles for related items

### LINKED ISSUE #203: The relatedItem generated by the system to reflect the relationship isContituentOf needs to dereference the object with the MODS title instead of the dc label
@blalbrit @LynnMcRae

### LINKED ISSUE #207: Dereferenced title for isConstituentOf relatedItem needs to be MODS too
We fixed this for collection references but missed it for isConstituentOf relationships when creating the relatedItem for the published MODS.  For relatedItem purposes we want" title : subtitle", though any common code involved should support just title, e.g., for facets.

John wanted me to mention some minor refactoring, like renaming "get_collection_title(obj)" which really fetches the title of any designated object. and making this usable by item-release when dereferencing isConstituentOf and collection for 856 tags.

### LINKED ISSUE #232: try to get rid of old usages of DC
as @ndushay pointed out just now, and as @LynnMcRae has pointed out in the past, we should try to avoid using DC values, and use MODS info instead.  here's an example where we still reference DC info:

https://github.com/sul-dlss/dor-services/blob/master/lib/dor/models/describable.rb#L185

``` ruby
  # load the title from the parent's DC.title
  titleInfo = doc.create_element 'titleInfo'
  title = doc.create_element 'title'
  title.content = parent_item.datastreams['DC'].title.first
  titleInfo << title
  relatedItem << titleInfo
```

cbeer:
Fixes #232 
Fixes #207 
Fixes #203

### DIFF
--- a/lib/dor/models/describable.rb
+++ b/lib/dor/models/describable.rb
@@ -178,7 +178,7 @@ module Dor
         # load the title from the parent's DC.title
         titleInfo = doc.create_element 'titleInfo'
         title = doc.create_element 'title'
-        title.content = parent_item.datastreams['DC'].title.first
+        title.content = Dor::Describable.get_collection_title(parent_item)
         titleInfo << title
         relatedItem << titleInfo
 
@@ -275,15 +275,11 @@ module Dor
     end
 
     def self.get_collection_title(obj)
-      xml = obj.descMetadata.ng_xml
-      title = ''
-      title_node = xml.at_xpath('//mods:mods/mods:titleInfo/mods:title', 'mods' => 'http://www.loc.gov/mods/v3')
-      if title_node
-        title = title_node.content
-        subtitle = xml.at_xpath('//mods:mods/mods:titleInfo/mods:subTitle', 'mods' => 'http://www.loc.gov/mods/v3')
-        title += " (#{subtitle.content})" if subtitle
-      end
-      title
+      obj.full_title
+    end
+
+    def full_title
+      stanford_mods.sw_title_display
     end
 
     private
--- a/lib/dor/models/identifiable.rb
+++ b/lib/dor/models/identifiable.rb
@@ -185,9 +185,7 @@ module Dor
     def get_related_obj_display_title(related_obj, default_title)
       return default_title unless related_obj
 
-      desc_md_ds = related_obj.datastreams['descMetadata']
-      desc_md_ds_title = desc_md_ds ? desc_md_ds.title_info.main_title.first : nil
-      desc_md_ds_title.present? ? desc_md_ds_title : default_title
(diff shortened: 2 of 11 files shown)


spec/dor/publishable_spec.rb
```diff
@@ -462,7 +463,10 @@ class ItemizableItem < ActiveFedora::Base
           # load child DC metadata fixture and set label
           dsid = 'DC'
           child_item.datastreams[dsid] = Dor::SimpleDublinCoreDs.from_xml read_fixture("#{druid}_#{dsid}.xml")
-          child_item.label = child_item.datastreams[dsid].title
+
```
cbeer: We ought to just use our normal fixture loader here..

lib/dor/models/identifiable.rb
```diff
@@ -185,9 +185,7 @@ def update_tag(old_tag, new_tag)
     def get_related_obj_display_title(related_obj, default_title)
       return default_title unless related_obj
 
-      desc_md_ds = related_obj.datastreams['descMetadata']
-      desc_md_ds_title = desc_md_ds ? desc_md_ds.title_info.main_title.first : nil
-      desc_md_ds_title.present? ? desc_md_ds_title : default_title
+      related_obj.full_title || default_title
```
drh-stanford: do we need to protect against a blank title here? previously it was using a `.present?` check...

lib/dor/models/describable.rb
```diff
@@ -275,15 +275,11 @@ def set_desc_metadata_using_label(force = false)
...
-        title = title_node.content
-        subtitle = xml.at_xpath('//mods:mods/mods:titleInfo/mods:subTitle', 'mods' => 'http://www.loc.gov/mods/v3')
-        title += " (#{subtitle.content})" if subtitle
-      end
-      title
+      obj.full_title
+    end
+
+    def full_title
+      stanford_mods.sw_title_display
```
drh-stanford: do we care that this is functionally different? that is, stanford_mods adds the part number/name to the title, and puts the subtitle in but not in parens. I'm guessing that stanford_mods is more correct anyway.

LynnMcRae:
"do we care that this is functionally different? that is, stanford_mods adds the part number/name to the title, and puts the subtitle in but not in parens. I'm guessing that stanford_mods is more correct anyway."  

correct, we'd prefer the current MODS rendering over the DC transformation from days gone by. In the large majority of cases, e.g. looked up collection titles, it will be the same.

lib/dor/models/identifiable.rb
cbeer: Stanford MODS handles that for us.

PR #268, opened 2016-11-15
add class hierarchy diagram to README

drh-stanford:


cbeer:
I should regenerate it since we're probably going to do away with `Dor::Base` again.

drh-stanford:
ok. closing.

PR #274, opened 2016-11-17
Memoize the fedora client certificate store to avoid unnecessary reinitialization

cbeer:
Depends on https://github.com/projecthydra/active_fedora/pull/1176

Fixes https://github.com/sul-dlss/dor_indexing_app/issues/109.

### DIFF
--- a/lib/dor/config.rb
+++ b/lib/dor/config.rb
@@ -131,6 +131,7 @@ module Dor
       connection_opts = { :url => fedora.safeurl, :user => fedora_uri.user, :password => fedora_uri.password }
       connection_opts[:ssl_client_cert] = OpenSSL::X509::Certificate.new(File.read(ssl.cert_file)) if ssl.cert_file.present?
       connection_opts[:ssl_client_key] = OpenSSL::PKey::RSA.new(File.read(ssl.key_file), ssl.key_pass) if ssl.key_file.present?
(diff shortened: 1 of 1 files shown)


lib/dor/config.rb
```diff
@@ -131,6 +131,7 @@ def fedora_config
       connection_opts = { :url => fedora.safeurl, :user => fedora_uri.user, :password => fedora_uri.password }
       connection_opts[:ssl_client_cert] = OpenSSL::X509::Certificate.new(File.read(ssl.cert_file)) if ssl.cert_file.present?
       connection_opts[:ssl_client_key] = OpenSSL::PKey::RSA.new(File.read(ssl.key_file), ssl.key_pass) if ssl.key_file.present?
+      connection_opts[:cert_store] = default_ssl_cert_store
```
drh-stanford: shouldn't this be `:ssl_cert_store`? see https://github.com/rest-client/rest-client/blob/adb990bd54596c76d5fa74d7f6d52d23e0cad321/lib/restclient/request.rb#L179-L184

cbeer: Good catch. I misread https://github.com/rest-client/rest-client/blob/adb990bd54596c76d5fa74d7f6d52d23e0cad321/lib/restclient/request.rb#L190.

PR #283, opened 2016-11-23
Refactor Describable#to_solr to avoid unnecessary or duplicative work

cbeer:


lib/dor/models/concerns/describable.rb
```diff
@@ -198,68 +200,74 @@ def add_constituent_relations(doc)
...
-        dc_doc.xpath('//dc:creator', dc: XMLNS_DC).each do |node|
-          creator = node.text
-        end
-        title = ''
-        dc_doc.xpath('//dc:title', dc: XMLNS_DC).each do |node|
-          title = node.text
+      mods_sources.each_pair do |meth, solr_keys|
+        vals = meth.is_a?(Array) ? stanford_mods.send(meth.shift, *meth) : stanford_mods.send(meth)
+
+        next if vals.nil? || (vals.respond_to?(:empty?) && vals.empty?)
```
drh-stanford: i see that this is a refactor for this set of conditions, but is this a problem? are blank values passed through? is that intentional? maybe add a comment for this as it's confusing, to me at least.

cbeer: I don't know what, if anything, the previous code was guarding against. I don't think it's blank values (or, if it is, the previous code wasn't guarding against that), but something bad coming back from stanford_mods?

PR #286, opened 2016-12-01
Stop generating dublin core fields for indexing

cbeer:
Now that Argo 3.10 is deployed to all environments, we should be able to stop indexing into these fields.

### DIFF
--- a/lib/dor/models/concerns/describable.rb
+++ b/lib/dor/models/concerns/describable.rb
@@ -201,34 +201,12 @@ module Dor
     def to_solr(solr_doc = {}, *args)
       solr_doc = super solr_doc, *args
       add_metadata_format_to_solr_doc(solr_doc)
-      add_dc_to_solr_doc(solr_doc)
       add_mods_to_solr_doc(solr_doc)
     end
 
     def add_metadata_format_to_solr_doc(solr_doc)
       solr_doc['metadata_format_ssim'] ||= []
       solr_doc['metadata_format_ssim'] += ['mods']
-    end
-
-    def add_dc_to_solr_doc(solr_doc)
-      dc_doc = generate_dublin_core(include_collection_as_related_item: false)
-      # we excluding the generated collection relation here; we instead get the collection
-      # title from Dor::Identifiable.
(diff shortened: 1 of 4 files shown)


peetucket:
Is there anyone else other than Argo that might care about these fields disappearing from the index?

cbeer:
I can't find any indication of that, no. I searched for a couple of the fields across the sul-dlss organization and found nothing.

peetucket:
discussed with @atz, double-checked sul-dlss/dor-fetcher-service, all seems fine

atz:
So basically the only thing the deleted code added was a questionable convenience field `creator_title_ssim`, combining title and creator.  That was a lot of work to build something we already have in the index, essentially.

cbeer:
And a variety of other `public_dc_*` fields.

LynnMcRae:
Chris, @blalbrit when we review #908 can we review the changes here as well.  Ben and I don't have an overview of all the changes and it's making me nervous about unintended side-effects, like coverage of terms for the quicksearch.

cbeer:
I'm not sure I know what you mean by "coverage of terms for the quicksearch"?

LynnMcRae:
words from subjects and abstract and titles and names and notes etc all participate in the keyword quicksearch I think, and I don't know if any of those values came exclusively from DC derived fields. There's so much redundancy in there it's hard to tell. it might be all good.

cbeer:
You're talking about the search box in argo? I reviewed the changes with @blalbrit the other week and we agreed we were covering the same metadata elements.

blalbrit:
@LynnMcRae @cbeer - I'm perfectly fine if we meet Wednesday to go over this. I don't think we're missing anything in terms of Argo search capabilities, but it's worthwhile doing a quick review. I'd also like to clarify the distinction between this PR and future work toward a cleaner (more readable) schema / set of field names. I see them as two different things and want to make sure we're not overloading this conversation with concerns and requirements that can be separated.

LynnMcRae:
sounds good.  One outcome is a nice documented list of specific changes we can make part of any notification, with perhaps solid information about which fields do what and why!

PR #287, opened 2016-12-01
Extract Dor::PublicDescMetadataService from Dor::Describable

cbeer:
This is pretty much a straight refactor of the existing code. The only functional change is an increased reliance on the public descriptive metadata for generating the public dc, instead of having two different paths for enhancing existing mods.


peetucket:
looks like extracting methods into a new class?

PR #293, opened 2016-12-02
Remove unused geoable methods

cbeer:
@drh-stanford I don't think these methods are being used and were just copied from `Dor::Describable`. As far as I can tell, `Dor::MetadataService` only retrieves MODS from Symphony, which I assume doesn't make sense to put into the `geoMetadata` datastream.


drh-stanford:
yes this is fine. i merged it.

PR #294, opened 2016-12-05
Use ActiveFedora finders in Dor::WorkflowObject.find_by_name

cbeer:


peetucket:
👍  from me but would prefer if someone else more knowledgable about activefedora has a look

PR #297, opened 2016-12-08
Remove unused configuration setter

cbeer:
I can't find any evidence of this being used. I'd welcome more eyes looking for uses, though.


peetucket:
I don't think I've ever worked with that datastream directly.  And yes, if used by the lyber-core base robot code, I'd expect an error at first use (doubtful there would be bad data though).

PR #298, opened 2016-12-08
Move ContentMetadataDs#public_xml into the Dor::PublicXmlService

cbeer:


lib/dor/services/public_xml_service.rb
```diff
@@ -11,42 +11,57 @@ def to_xml
...
-      pub.add_child(object.datastreams['rightsMetadata'].ng_xml.root.clone)
-
-      rels = object.public_relationships.root
-      pub.add_child(rels.clone) unless rels.nil? # TODO: Should never be nil in practice; working around an ActiveFedora quirk for testing
-      pub.add_child(object.generate_dublin_core.root.clone)
-      pub.add_child(Nokogiri::XML(object.generate_public_desc_md).root.clone)
-      pub.add_child(release_xml.root.clone) unless release_xml.xpath('//release').children.size == 0 # If there are no release_tags, this prevents an empty <releaseData/> from being added
+      pub.add_child(public_rights_metadata.root)
+      pub.add_child(public_relationships.root) unless public_relationships.nil? # TODO: Should never be nil in practice; working around an ActiveFedora quirk for testing
+      pub.add_child(object.generate_dublin_core.root)
```
jmartin-sul: is it that `generate_dublin_core` produces a new XML element anyway, and so doesn't need to be `clone`d?

lib/dor/services/public_xml_service.rb
```diff
@@ -11,42 +11,94 @@ def to_xml
...
-      pub.add_child(public_content_metadata.root.clone) if public_content_metadata.xpath('//resource').any?
-      pub.add_child(object.datastreams['rightsMetadata'].ng_xml.root.clone)
-
-      rels = object.public_relationships.root
-      pub.add_child(rels.clone) unless rels.nil? # TODO: Should never be nil in practice; working around an ActiveFedora quirk for testing
-      pub.add_child(object.generate_dublin_core.root.clone)
-      pub.add_child(Nokogiri::XML(object.generate_public_desc_md).root.clone)
-      pub.add_child(release_xml.root.clone) unless release_xml.xpath('//release').children.size == 0 # If there are no release_tags, this prevents an empty <releaseData/> from being added
+      pub.add_child(public_relationships.root) unless public_relationships.nil? # TODO: Should never be nil in practice; working around an ActiveFedora quirk for testing
+      pub.add_child(object.generate_dublin_core.root)
```
jmartin-sul: i had left my question in the first commit, the line got moved in the second commit, so github shows the comment as being "outdated", but question still applies here, i think.

cbeer: Yes, exactly. It's generated fresh on every call.

PR #299, opened 2016-12-08
[WIP] consider skipped as a completed state

drh-stanford:
@lmcglohon @peetucket This PR fixes a problem with `to_solr` that produces incorrect facets for the `wf_xxx_ssim` fields.

For example,

```
"releaseWF:release-publish",
"releaseWF:release-publish:waiting",
"releaseWF:release-publish:ready"
```

should be

```
"releaseWF:release-publish",
"releaseWF:release-publish:completed"
```

There are no existing tests that deal with the state computations, and as far as I can tell, no existing specs on the computation of the workflow facets.

### DIFF
--- a/lib/dor/workflow/process.rb
+++ b/lib/dor/workflow/process.rb
@@ -43,7 +43,7 @@ module Workflow
     def note          ; @attrs['note']         ; end
     def version       ; @attrs['version']      ; end
     def priority      ; @attrs['priority']     ; end
-    def completed?    ; status == 'completed'  ; end
+    def completed?    ; %w(completed skipped).include?(status); end
     def error?        ; status == 'error'      ; end
     def waiting?      ; status == 'waiting'    ; end
     def date_time     ; @attrs['datetime']     ; end


lmcglohon:
@drh-stanford @peetucket Please do not accept and/or merge this PR until you talk with me

jmartin-sul:
i'm not sure if this is actually work in progress in the sense of code changes left to be made... but i edited the title so that someone wouldn't jump in and accidentally merge it based on the approval, in case they miss @lmcglohon's comment.

drh-stanford:
I'm closing this PR as my understanding now is that there's a functional requirement to see `skipped` as a state in Argo, so the `to_solr` method will need an approach different from this PR to implement that....

LynnMcRae:
Given the wording of the ticket subject ("consider skipped as a completed") was there a misunderstanding that "skipped" should somehow be translated/expressed as "completed"?  The only sense in which skipped=completed is that they both satisfy
 - the workflow pre-requisite notion of done-ness, and
 - the workflow archiving logic to move a "completed" WF instance from the active to archive table

In both cases this is accomplished by means of "status is completed or skipped" expressed in their internal SQL queries.

In all other cases "skipped" is its own status to be proudly expressed as recorded; the index/facets can be simple concatenation of values found in the workflows. That would be true even if someone recorded a bogus status, like the occasional "hold" we've used to pause a workflow (any non-recognized value will do), or a hand-entered "complete" instead of "completed".

drh-stanford:
Yes, I think there was a misunderstanding on my part. 

The specific problem I saw was that  "skipped" is recorded in the `wf_xxx_ssim` facets incorrectly:

```
"releaseWF:release-publish",
"releaseWF:release-publish:waiting",
"releaseWF:release-publish:ready"
```

instead of

```
"releaseWF:release-publish",
"releaseWF:release-publish:skipped",
```

See https://github.com/sul-dlss/dor-services/blob/master/lib/dor/workflow/process.rb#L55-L57 for what I assumed is incorrect logic since it doesn't consider "skipped" as meeting the prereqs.

*BUT*, I cannot find the example I had :( and I didn't put it in the ticket. 

So it may not be a real problem and just confusion on my part. Unless we find an example of something that is skipped but recorded in the facet incorrectly, then I suggest we can put this issue to rest.

LynnMcRae:
If the next step in the item-release workflow never runs because the preceding step (release-publish) is marked skipped instead of completed, then we do have a problem. However, I thought the robot framework / common code was evaluating prerequisites to come up with the "ready" status, so everyone would respond to completed or skipped the same. If this is the process.rb code you cite, then skipped=completed for prerequisites does seem to be missing and we would want your changes in https://github.com/sul-dlss/dor-services/commit/6edf70737515b67dcea144cf562705b7b269176b but only for the purpose of prerequisite evauluation? I am not confident in my code sleuthing, so take it with a measure of salt.

drh-stanford:
I created a ticket https://github.com/sul-dlss/dor-services/issues/300 to continue this discussion. I framed it as an audit and I dug into the workflow service, workflow archiver, and workflow gem and they all seem to handle the `skipped` status correctly.

PR #304, opened 2017-01-18
registration_service_spec: add specs for #create_from_request

ndushay:
do not merge - this is morphing!

Adding these specs to 'document' that source_id must have exactly one colon -- this is a problem encountered by the was-registrar rails app when trying to register an object.  

(per conversation with @eefahy on "how do we sustain-ably document the API being used by other apps?" in the context of doing better monitoring and testing robots.)


ndushay:
closed b/c diff solution taken, per #306

PR #309, opened 2017-01-25
When publishing metadata, use #content instead of #to_xml

cbeer:
for compatibility with all datastreams, including datastreams that are not declared as part of the model for a class.


cbeer:
I'm not sure it's feasible/useful to try to construct a test for the hydrus-collection-with-contentMetadata-even-if-it-isnt-declared-in-the-model. I find myself fighting hard against ActiveFedora to the point where everything is just all stubbed out anyway.

PR #310, opened 2017-02-15
add rights type codes for additional locations

jmartin-sul:
addresses sul-dlss/argo#934

### DIFF
--- a/lib/dor/datastreams/rights_metadata_ds.rb
+++ b/lib/dor/datastreams/rights_metadata_ds.rb
@@ -54,8 +54,12 @@ module Dor
       'world-nd' => 'World (no-download)',
       'stanford' => 'Stanford',
       'stanford-nd' => 'Stanford (no-download)',
-      'loc:spec' => 'Location spec',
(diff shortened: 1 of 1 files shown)


jmartin-sul:
@LynnMcRae 

here are how a few different versions look in the registration drop down (all with the same "Location" prefix in the human readable text):
screen shot 2017-02-15 at 8 46 10 am
screen shot 2017-02-15 at 8 47 40 am
screen shot 2017-02-15 at 8 49 04 am


the underlying location code that's actually stored in the xml is the same regardless, so easy enough to change the human readable version later if needed.  i think i'm slightly partial to option 2, unless the people doing the registering find it more convenient/readable to go by short codes, in which case i'd be partial to option 3.

LynnMcRae:
The middle "Location: name" version is lovely, let's go with that. Drop the "reading room" and "listening room" qualifiers to the longer names however. Thanks.

jmartin-sul:
@peetucket, travis is being slow, but this should be ready for review if/when the tests pass.

i believe the existing tests cover these new entries.

PR #313, opened 2017-02-22
descMetadata mods header, upgrade to 3.6

jmartin-sul:
addresses sul-dlss/argo#942


spec/models/concerns/describable_spec.rb
```diff
@@ -125,10 +125,11 @@ class SimpleItem < ActiveFedora::Base
       expect{@obj.set_desc_metadata_using_label}.to raise_error(StandardError)
     end
     it 'should run if there is content in the descriptive metadata stream and force is true' do
-      @obj.set_desc_metadata_using_label(false)
+      allow(@obj.descMetadata).to receive(:new?).and_return(false)
+      @obj.set_desc_metadata_using_label(true)
```
jmartin-sul: this wasn't related to the PR, just an incidental change to make the test do what it was claiming to do in the first place

PR #321, opened 2017-04-24
Fix tech metadata for non items

### LINKED ISSUE #320: Tech-metadata robot fails when called on non-object types
Tech-metadata generation depends on methods in the `Itemizable` concern, which is unavailable to other object types.

peetucket:
fixes #320 

also addresses another issue if an exception is raised in sdr-ingest

### DIFF
--- a/spec/models/concerns/preservable_spec.rb
+++ b/spec/models/concerns/preservable_spec.rb
@@ -49,8 +49,15 @@ describe Dor::Preservable do
 
   end
 
-  it 'builds the technicalMetadata datastream' do
+  it 'builds the technicalMetadata datastream if the object is an item' do
+    allow(item).to receive(:is_a?).with(Dor::Item).and_return(true)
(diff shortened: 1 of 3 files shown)


spec/models/concerns/preservable_spec.rb
```diff
@@ -49,11 +49,18 @@ class PreservableItem < ActiveFedora::Base
 
   end
 
-  it 'builds the technicalMetadata datastream' do
+  it 'builds the technicalMetadata datastream if the object is an item' do
+    allow(item).to receive(:is_a?).and_return(Dor::Item)
```
jmartin-sul: i think `is_a?` is supposed to return a boolean:
http://ruby-doc.org/core-2.4.1/Object.html#method-i-is_a-3F

peetucket: good catch -- fixing both tests, rebasing and pushing in a minute

PR #322, opened 2017-06-30
Index the mime types for files

cbeer:


ndushay:
LGTM;  not sure if there's any "keeping things in sync" that needs to be addressed before a merge.

cbeer:
Not that I know of.

PR #328, opened 2017-07-27
Move connection creation out of the configuration

jcoyne:
Classes should only have one responsibility.

This deprecates `Dor::Config#make_rest_client` and
`Dor::Config#autoconfigure`


cbeer:
👍 my cursory search of sul-dlss projects suggests `autoconfigure` and `make_rest_client` are entirely unused.

PR #331, opened 2017-08-01
Remove object content from stacks when rights are set to dark

tingulfsen:
Fixes sul-dlss/argo#977

### DIFF
--- a/lib/dor/models/concerns/governable.rb
+++ b/lib/dor/models/concerns/governable.rb
@@ -32,6 +32,20 @@ module Dor
 
     def set_read_rights(rights)
       rightsMetadata.set_read_rights(rights)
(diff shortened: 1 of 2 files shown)


ndushay:
LGTM; 👍  to merge.

jmartin-sul:
sorry, this just got merged, i was late leaving a review, but the one suggestion i had was:  what about putting `unshelve_and_unpublish` in `Contentable`, or `ContentMetadataDS`?  that'd require a bit more conditional logic in `set_read_rights`, but it also feels like a slightly more appropriate home.  it feels a little odd to me to manipulate `contentMetadata` in `Governable`.

that was the only feedback i had.  functionally, looks totally fine, and that's not a change i would've pushed very strongly for if others disagreed or thought it wasn't worth the effort.

PR #336, opened 2017-11-13
Fix spelling of decommission in method names

cbeer:


blalbrit:
thank you thank you thank you.

lib/dor/services/merge_service.rb
```diff
@@ -58,23 +58,25 @@ def copy_workspace_content
...
           Dor::CleanupService.cleanup_by_druid @current_secondary.pid
           Dor::Config.workflow.client.archive_active_workflow 'dor', @current_secondary.pid
         rescue => e
-          @logger.error "Unable to decomission #{@current_secondary.pid} with primary object #{@primary.pid}: #{e.inspect}"
+          @logger.error "Unable to decommission #{@current_secondary.pid} with primary object #{@primary.pid}: #{e.inspect}"
           @logger.error e.backtrace.join("\n")
         end
       end
     end
+    alias_method :decomission_secondaries, :decommission_secondaries=
```
jcoyne: I don't see where `decommission_secondaries=` is defined? should the `=` be dropped?

cbeer: Yep, it was the same copy-pasta as the other method.

PR #338, opened 2018-02-06
Add support for a @role attribute on a contentMetadata's file element

cbeer:
Related to https://github.com/sul-dlss/content_search/issues/47.

For OCR files, we're strongly considering adding a role="transcription" attribute to the <file> element that references the OCR'ed (or manually transcribed, or speech-to-text'ed, etc) content. This PR makes it possible to pass this attribute through our stack.

The `@role` attribute was spec'ed, but never implemented, and was initially described as:

> Possible attribute to distinguish like files in a resource where a published resource wants alternate representations, e.f., cropped/uncropped for a single resource.

### DIFF
--- a/lib/dor/datastreams/content_metadata_ds.rb
+++ b/lib/dor/datastreams/content_metadata_ds.rb
@@ -19,6 +19,7 @@ module Dor
           t.mimeType :path => { :attribute => 'mimeType' }, :index_as => [:displayable]
           t.dataType :path => { :attribute => 'dataType' }, :index_as => [:displayable]
           t.size     :path => { :attribute => 'size'     }, :index_as => [:displayable]    # , :data_type => :long
+          t.role     :path => { :attribute => 'role' },     :index_as => [:not_searchable]
           t.shelve   :path => { :attribute => 'shelve'   }, :index_as => [:not_searchable] # , :data_type => :boolean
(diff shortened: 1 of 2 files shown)


lib/dor/datastreams/content_metadata_ds.rb
```diff
@@ -95,10 +97,12 @@ def to_solr(solr_doc = {}, *args)
...
           preserved_size += file['size'].to_i if file['preserve'] == 'yes'
           shelved_size += file['size'].to_i if file['shelve'] == 'yes'
-          next unless file['shelve'] == 'yes'
-          counts['shelved_file'] += 1
-          first_shelved_image ||= file['id'] if file['id'] =~ /jp2$/
+          if file['shelve'] == 'yes'
+            counts['shelved_file'] += 1
+            first_shelved_image ||= file['id'] if file['id'] =~ /jp2$/
+          end
           mime_types << file['mimetype']
```
jkeck: It looks like we were skipping adding the file's mimetype to the `mime_types` set before this change.  Is it an intentional change to add un-shelved file's mimetypes here?

cbeer: Yes; I caught this when making the change. It's in a separate commit (3a50dd8) than the `@role` attribute work.

jkeck: Ah, didn't notice the separate commit.  Thanks.

PR #340, opened 2018-03-01
Fix outdated call to #next_pid in SuriService

mjgiarlo:
If `Config.suri.mint_ids` is false, the SuriService raises a NoMethodError calling `#next_pid` on a Rubydora::Repository instance. Rubydora::Repository instances do not define this method; the method is defined on Rubydora::RestApiClient which can be accessed via `#api` on the repository.

Additionally, the xpath that is used in the same `else` clause no longer parses pids out of the `pidList`. I have added code that tries a second XPath if the first fails.

This bug has been hidden for 6ish years due to how the spec is mocking things out.

I made these changes so the latest hydra_etd work can use dor-services without patching around the SuriService.

### DIFF
--- a/lib/dor/services/suri_service.rb
+++ b/lib/dor/services/suri_service.rb
@@ -17,8 +17,14 @@ module Dor
         ids = resource["identifiers?quantity=#{quantity}"].post('').chomp.split(/\n/).collect { |id| "#{Config.suri.id_namespace}:#{id.strip}" }
       else
         repo = ActiveFedora::Base.respond_to?(:connection_for_pid) ? ActiveFedora::Base.connection_for_pid(0) : ActiveFedora.fedora.connection
-        resp = Nokogiri::XML(repo.next_pid(numPIDs: quantity))
+        resp = Nokogiri::XML(repo.api.next_pid(numPIDs: quantity))
         ids = resp.xpath('/pidList/pid').collect { |node| node.text }
+        # With modernish (circa 2015/6) dependencies, including Nokogiri and
+        # ActiveFedora/Rubydora, `ids` is `[]` above. If that is the case, try
+        # the XPath that works (confirmed with most recent `hydra_etd` work)
+        if ids.empty? && resp.root.namespaces.any?
+          ids = resp.xpath('/xmlns:pidList/xmlns:pid').collect { |node| node.text }
+        end
       end
       want_array ? ids : ids.first
 
--- a/spec/services/registration_service_spec.rb
+++ b/spec/services/registration_service_spec.rb
@@ -6,7 +6,7 @@ describe Dor::RegistrationService do
 
   before :each do
     @pid = 'druid:ab123cd4567'
-    @mock_repo = double(Rubydora::Repository, :url => 'foo')
+    @mock_repo = instance_double(Rubydora::Repository)
     @apo = instantiate_fixture('druid:fg890hi1234', Dor::AdminPolicyObject)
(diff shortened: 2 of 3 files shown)


spec/services/suri_service_spec.rb
```diff
@@ -52,6 +52,9 @@
 
     before :each do
       @mock_repo = double(Rubydora::Repository)
+      @mock_client = double(Rubydora::RestApiClient)
```
jmartin-sul: what about turning these into `instance_double` calls, instead of regular `double` calls?  would that have caught the problem with the bad mocking before, since `instance_double` requires `expect`/`allow` mocked method calls to exist on the class being mocked?

lib/dor/services/suri_service.rb
```diff
@@ -17,8 +17,14 @@ def self.mint_id(quantity = nil)
...
       else
         repo = ActiveFedora::Base.respond_to?(:connection_for_pid) ? ActiveFedora::Base.connection_for_pid(0) : ActiveFedora.fedora.connection
-        resp = Nokogiri::XML(repo.next_pid(numPIDs: quantity))
+        resp = Nokogiri::XML(repo.api.next_pid(numPIDs: quantity))
         ids = resp.xpath('/pidList/pid').collect { |node| node.text }
+        # With modernish (circa 2015/6) dependencies, including Nokogiri and
+        # ActiveFedora/Rubydora, `ids` is `[]` above. If that is the case, try
+        # the XPath that works (confirmed with most recent `hydra_etd` work)
+        if ids.empty? && resp.root.namespaces.any?
+          ids = resp.xpath('//xmlns:pid').collect { |node| node.text }
```
ndushay: so this is ... any nodes in the pid namespace?  zounds that seems fraught with ickies.  Might there be particular elements in scope?  Might this xml get big ('//xxx' xpath expressions are expensive, but it doesn't matter with tiny xml).

mjgiarlo: @ndushay The XML is tiny. Fedora 3, which we talk to via Rubydora, allows you to request 1 or more PIDs, and those are returned in an XML blurb that looks like the following, e.g. if you asked for 3 PIDs:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<pidList namespaces="here">
  <pid>druid:123</pid>
  <pid>druid:124</pid>
  <pid>druid:125</pid>
</pidList>
```

You can see the XML generation code in Fedora 3 here: https://github.com/fcrepo3/fcrepo/blob/9c41b09a21a3a2615791fa4c614095a14512940f/fcrepo-server/src/main/java/org/fcrepo/server/rest/DefaultSerializer.java#L65-L84

mjgiarlo: @ndushay I will change the XPath from `//xmlns:pid` to `/xmlns:pidList/xmlns:pid` to more closely mirror the code already in `Dor::SuriService`.

spec/services/registration_service_spec.rb
```diff
@@ -6,7 +6,7 @@
 
   before :each do
     @pid = 'druid:ab123cd4567'
-    @mock_repo = double(Rubydora::Repository, :url => 'foo')
+    @mock_repo = instance_double(Rubydora::Repository)
```
mjgiarlo: I tossed this in for free.

mjgiarlo:
@ndushay @jmartin-sul Thanks for the reviews! Can I ask y'all to re-review?

PR #345, opened 2018-03-14
Allow newer versions of rake

jcoyne:
per http://guides.rubygems.org/patterns/#semantic-versioning

> It’s also important to know that if you specify a major version only, like this:

```
spec.add_runtime_dependency 'library', '~> 2'
```
> It will only use the latest version from the 2.x series – so 2.3.0 – and not 3.0.0.


atz:
`This behaviour may surprise some people, but automatically allowing any major version past version 2 is more surprising behaviour.`

Why would that be more surprising to somebody who uses the squiggly operator in all the other cases where it behaves differently?  Are they writing a DSL for people who can't understand the basis of the DSL or for people who just *really* like exceptions?

Ugh, anyway, merging...

PR #346, opened 2018-03-19
Explicitly set the defType to lucene for a lucene query

cbeer:


jcoyne:
Is this fixing a bug with Solr 7.2? https://lucene.apache.org/solr/7_2_0/changes/Changes.html#v7.2.0.upgrade_notes

cbeer:
Yes, exactly. Although, even still, we shouldn't send half-baked queries to request handlers and hope they just work.

PR #349, opened 2018-03-21
Dependency updates

atz:
- `lyber-utils` not used
- let `druid-tools` use contemporary releases
- bump `moab-versioning`


atz:
@jcoyne unaware of gotchas.  `moab-versioning` has a pretty tame domain w/o a lot of volatility in functionality (mostly in specs).  `2.x` is only 5 releases ago with `3.0.0` only cut in November of 2017.  

Obviously we will need a `dor-services` release at some point to unblock our dependency issue.  For my money, it's a minor release.

atz:
@jcoyne are there any intervening changes here since last release that warrant a major version?

PR #359, opened 2018-05-16
Deprecate unused thumb_url and supporting methods

jcoyne:


lib/dor/models/concerns/publishable.rb
```diff
@@ -39,6 +41,7 @@ def thumb
 
        thumb_image
     end
+    deprecation_deprecate :thumb
```
peetucket: `thumb` is still used in generating public XML here: https://github.com/sul-dlss/dor-services/blob/master/lib/dor/services/public_xml_service.rb#L25

peetucket: originally here when first implemented: https://github.com/sul-dlss/dor-services/pull/237/files#diff-831fbf194eaba823278738aa38837ceaR100 (e

peetucket:
If `#thumb` is still being used in the public XML, doesn't leaving it mark as deprecated risk that some future developer will inadvertently remove it without realizing this?

jcoyne:
@peetucket I updated `PublicXmlService` to use the new`ThumbnailService` and not `object.thumb`

peetucket:
right, never mind, i see now the extracted service method is not deprecated 👍

PR #360, opened 2018-05-17
Assign a PID when saving rather than on initialization

jcoyne:
This will allow us to do:

```ruby
form_for Dor::Item.new
```
without calling SURI to mint a pid

### DIFF
--- a/lib/dor/models/concerns/identifiable.rb
+++ b/lib/dor/models/concerns/identifiable.rb
@@ -18,13 +18,13 @@ module Dor
         @object_type = str
         Dor.registered_classes[str] = self
       end
-    end
 
-    def initialize(attrs = {})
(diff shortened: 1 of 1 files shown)


mjgiarlo:
@jmartin-sul do you still want to wait on a review from @atz?

jmartin-sul:
@mjgiarlo, i've got no objection to merging this.  it wasn't something i felt comfortable reviewing myself, and i only requested @atz's review because i know he knows `ActiveFedora` ins and outs better than i do.  but if you're comfortable with this approach from having used it elsewhere, that's good enough for me.

mjgiarlo:
@jmartin-sul OK! Thanks, John. Wanted to hear from you before I acted on this PR. :beers:

PR #362, opened 2018-06-04
Add build for ruby 2.3.4

tallenaz:
prep for ruby upgrade on sul-robots*-test
  dor-services is on sul-robots1-test


jcoyne:
@tallenaz can you say something about why you are upgrading to Ruby 2.3.4 and not 2.3.7?

tallenaz:
@jcoyne @aaron-collier and I agreed on testing 2.3.4. I'd be fine with 2.3.7 if Aaron agrees.

aaron-collier:
@jcoyne 2.3.7 may be fine, I believe I was going based off of list of versions in common-accessioning, which lists 2.3.4, so if 2.3.7 is OK too, I'm fine with it.

PR #364, opened 2018-06-05
Send HTTP API notifications to the purl services endpoint (a.k.a. pur…

cbeer:
…l-fetcher) when publishing metadata

### DIFF
--- a/lib/dor/models/concerns/publishable.rb
+++ b/lib/dor/models/concerns/publishable.rb
@@ -102,6 +102,7 @@ module Dor
       else
         # Clear out the document cache for this item
         DigitalStacksService.prune_purl_dir pid
+        publish_delete_on_success
       end
     end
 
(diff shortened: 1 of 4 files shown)


jcoyne:
This looks good.  If I understand correctly, we are telling purl-fetcher that we are doing delete, but we're notifying discovery-dispatcher of deletes in the old way (putting a file in the nfs mount).   Is there a road map here?  Will we be able to turn off the old way when some condition is met?

cbeer:
Right; my roadmap is:

- get a version of dor-services out that supports notifying purl-fetcher over HTTP, while still maintaining the current approach for complete backwards compatibility
- add the appropriate configuration to start notifying purl-fetcher over HTTP to the services we believe need it (argo, robots, dor-services-app?)
- evaluate while making sure things go smoothly
- drop the legacy code path
- wait to see if any other services are writing to those NFS "endpoints"; throw them away and unmount them.

lib/dor/models/concerns/publishable.rb
```diff
@@ -114,19 +115,37 @@ def publish_metadata_remotely
     end
 
     ##
-    # When publishing a PURL, we drop a `aa11bb2222` file into the `local_recent_changes` folder
+    # When publishing a PURL, we notify purl-fetcher of changes.
+    # We (for legacy reasons) also drop a `aa11bb2222` file into the `local_recent_changes` folder
```
jcoyne: This doc needs fixing.  The "also" no longer applies.

cmh2166:
just a fyi, this work will also come up again in the sdr3 / general architecture discussions to extend the roadmap to a more palatable replacement, but this is a good path for now. thanks @cbeer for doing this work to try to simplify where possible our current infra.

PR #379, opened 2018-11-15
Remove graph generation from dor-services

jcoyne:
It is not a shared concern, it only ever applied to Argo, and we're removing it from there.

This should be considered a breaking change, so it should go into a major release.
https://github.com/sul-dlss/argo/pull/1179


mjgiarlo:
:red_circle:

PR #420, opened 2018-12-10
Delegate has_model to the resource

jcoyne:


jcoyne:
depends on https://github.com/samvera/active_fedora/pull/1374 being merged/released

PR #424, opened 2018-12-19
Deprecate Eventable#add_event

jcoyne:


justinlittman:
Can you point me to where `events.add_event()` is?

jcoyne:
@justinlittman https://github.com/sul-dlss/dor-services/blob/master/lib/dor/datastreams/events_ds.rb#L33

PR #436, opened 2018-12-21
Deprecate generate_public_desc_md

### LINKED ISSUE #431: Deprecate generate_public_desc_md

jcoyne:
Refactor code such that we don't generate xml with nokogiri and then
reparse it back into a nokogiri document.

Fixes #431


justinlittman:
Approve pending conflict fix.

PR #444, opened 2018-12-21
Extract ProcessService from Processable

### LINKED ISSUE #432: extract status service from Processable

jcoyne:
Fixes #432


spec/models/workflow_object_spec.rb
```diff
@@ -15,17 +15,18 @@
     end
   end
 
+  # TODO: Move to the DataIndexer spec
```
justinlittman: Does this TODO need to be done?

jcoyne: It was not my intention to do it in this PR

PR #448, opened 2018-12-21
Deprecate Item#public_xml

jcoyne:


justinlittman:
Approve pending happy Travis.

PR #455, opened 2018-12-21
XML munging should be encapsulated by the datastream objects

jcoyne:


spec/datastreams/default_object_rights_ds_spec.rb
```diff
@@ -32,7 +32,7 @@
   it 'understands terms of a rightsMetadata and will normalize the text' do
     subject.normalize!
     expect(subject.copyright).to eq(['All rights reserved.'])
-    expect(subject.use_statement).to eq(['You may re-distribute this object, unaltered, with attribution to the author.'])
+    expect(subject.use_statement).to eq 'You may re-distribute this object, unaltered, with attribution to the author.'
```
mjgiarlo: Why did this change? Do we care?

jcoyne: Because we only ever use as a single valued field, so we moved up that `first` call into the datastream.

PR #468, opened 2019-01-07
Restore and deprecate Dor::Processable::STEPS. Its used by Argo

jcoyne:


lib/dor/models/concerns/processable.rb
```diff
@@ -16,6 +16,8 @@ module Processable
                    autocreate: true
     end
 
+    STEPS = ActiveSupport::Deprecation::DeprecatedConstantProxy.new('STATUS', 'Dor::StatusService::STEPS')
```
mjgiarlo: Why is the first arg `STATUS`?

jcoyne: Ah, typo. Thanks.

PR #519, opened 2019-03-15
Simplifies registering metadata handlers.

### LINKED ISSUE #517: get rid of Dor::MetadataService.register
This is never used outside of dor-services and it makes it harder to read the code.

justinlittman:
closes #517

### DIFF
--- a/lib/dor/services/metadata_service.rb
+++ b/lib/dor/services/metadata_service.rb
@@ -1,36 +1,14 @@
 # frozen_string_literal: true
 
 require 'cache'
+require 'dor/services/metadata_handlers/catalog_handler'
 
 module Dor
   class MetadataError < RuntimeError; end
 
-  #  class MetadataHandler
-  #
-  #    def fetch(prefix, identifier)
-  #      ### Return metadata for prefix/identifier combo
-  #    end
-  #
-  #    def label(metadata)
-  #      ### Return a Fedora-compatible label from the metadata format returned by #fetch
-  #    end
-  #
-  #  end
-
   class MetadataService
     class << self
       @@cache = Cache.new(nil, nil, 250, 300)
-
-      def register(handler_class)
-        %w(fetch label prefixes).each do |method|
-          raise TypeError, "Metadata handlers must define ##{method}" unless handler_class.instance_methods.include?(method) || handler_class.instance_methods.include?(method.to_sym)
-        end
-        handler = handler_class.new
-        handler.prefixes.each do |prefix|
-          handlers[prefix.to_sym] = handler
-        end
-        handler
-      end
 
       def known_prefixes
         handlers.keys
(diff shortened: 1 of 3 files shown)


lib/dor/services/metadata_service.rb
```diff
@@ -69,13 +70,21 @@ def handler_for(prefix)
...
       private
 
+      # def handlers
+      #   @handlers ||= {}
+      # end
       def handlers
-        @handlers ||= {}
+        @handlers ||= {}.tap do |handlers|
+          # There's only one. If additional handlers are added, will need to be registered here.
+          register(CatalogHandler.new)
```
jcoyne: I'd even get rid of the `register` logic and just return the `CatalogHandler` instance here.

lib/dor/services/metadata_service.rb
```diff
@@ -69,13 +70,21 @@ def handler_for(prefix)
 
       private
 
+      # def handlers
+      #   @handlers ||= {}
+      # end
       def handlers
-        @handlers ||= {}
+        @handlers ||= {}.tap do |handlers|
```
jcoyne: This inner variable is never used?

lib/dor/services/metadata_service.rb
```diff
@@ -21,16 +22,16 @@ class MetadataService
...
-        %w(fetch label prefixes).each do |method|
-          raise TypeError, "Metadata handlers must define ##{method}" unless handler_class.instance_methods.include?(method) || handler_class.instance_methods.include?(method.to_sym)
-        end
-        handler = handler_class.new
-        handler.prefixes.each do |prefix|
-          handlers[prefix.to_sym] = handler
-        end
-        handler
-      end
+      # def register(handler_class)
```
jcoyne: Can this stuff be removed?

lib/dor/services/metadata_service.rb
jcoyne: Would you mind calling the variable `handlers` something different than the method name?

lib/dor/services/metadata_service.rb
```diff
@@ -62,20 +63,25 @@ def label_for(identifier)
...
 
       def handlers
-        @handlers ||= {}
+        @handlers ||= {}.tap do |handlers|
+          # There's only one. If additional handlers are added, will need to be registered here.
+          register(CatalogHandler.new, handlers)
+        end
+      end
+
+      def register(handler, handlers)
```
jcoyne: would you mind renaming `handlers` so as not to confuse it with the method of the same name.

justinlittman: So at least on line 73, rubocop insists:
```

File: lib/dor/services/metadata_service.rb [hunk 5]
```diff
@@ -62,20 +40,25 @@ def label_for(identifier)
...
+        raise MetadataError, "Unknown metadata prefix: #{prefix}" if handler.nil?
 
         handler
       end
 
       private
 
       def handlers
-        @handlers ||= {}
+        @handlers ||= {}.tap do |md_handlers|
```
jcoyne: do we need `{}.tap ...` could this just be?
```ruby
@handlers ||= register(CatalogHandler.new, {})
```

jcoyne: I guess the tap makes sense if the idea is to encourage more `register()` lines here.

PR #532, opened 2019-04-09
Deprecate Dor.load_instance

### LINKED ISSUE #528: deprecate load_instance
we can just use `Dor.find`

jcoyne:
Fix #528


spec/services/indexing_service_spec.rb
```diff
@@ -136,42 +136,42 @@
     end
 
     it 'handles old, primitive arguments' do
-      expect(Dor).to receive(:load_instance).with(@mock_pid).and_raise(ActiveFedora::ObjectNotFoundError)
+      expect(Dor).to receive(::find).with(@mock_pid).and_raise(ActiveFedora::ObjectNotFoundError)
```
justinlittman: `::`?

jcoyne: that was a find/replace error.

PR #539, opened 2019-04-11
Deprecate reindex_pid_remotely

jcoyne:
Ref #535


jcoyne:
#535 explains that it’s moving to Argo

PR #544, opened 2019-04-11
Remove SdrIngestService

jcoyne:
it moved to common-accessioning

Ref https://github.com/sul-dlss/common-accessioning/pull/209


justinlittman:
Approve pending resolving conflict.

PR #578, opened 2019-04-18
Move knowledge about how to retrieve the rights to the datastream

jcoyne:


jcoyne:
@justinlittman I don't believe those cases could ever occur. There is no way to have an object that responds to `rights` but does not respond to `rightsMetadata`.  There is no way for the `rightsMetadata` method to ever return `nil`.

PR #581, opened 2019-04-22
Move the Etd/Part model in from etd_models

jcoyne:
This will make it easier as all models in DOR will be in one codebase


mjgiarlo:
@jcoyne Conflict needs resolving.

PR #586, opened 2019-04-23
Move methods that are about the object to Dor::Abstract

jcoyne:


mjgiarlo:
:red_circle:

jcoyne:
Yeah, this really needed this second commit to be complete: #588

PR #600, opened 2019-04-23
Merge Geoable concern into Item

jcoyne:
Because its only used once and this is less indirection


mjgiarlo:
@jcoyne Approved. Needs conflict resolved before being merged.

PR #601, opened 2019-04-23
Extract an EmbargoService

jcoyne:


mjgiarlo:
@jcoyne Approved but conflict needs resolving first.

PR #602, opened 2019-04-23
Extract PublishedRelationshipsFilter from the model

jcoyne:


jcoyne:
The test is in ./spec/services/public_desc_metadata_service_spec.rb.  Should I move it?

jcoyne:
I made a more focused test just for this class.

PR #630, opened 2019-05-22
Change to a static config and drop confstruct

jcoyne:
Confstruct is not well maintained and it's 'accept anything' API can lead to misconfiguration


spec/spec_helper.rb
```diff
@@ -29,19 +29,18 @@
...
+      local_workspace_root: File.join(this.fixture_dir, 'workspace'),
+      local_stacks_root: File.join(this.fixture_dir, 'stacks'),
+      local_document_cache_root: File.join(this.fixture_dir, 'purl')
+    )
     allow(ActiveFedora).to receive(:fedora).and_return(double('frepo').as_null_object) # must be used in per-request context: :each not :all
   end
 
   def unstub_config
-    Dor::Config.pop!
+    puts 'this does nothning'
```
mjgiarlo: typo: `nothning`

PR #646, opened 2019-07-11
show the object as accessioned if it has an accesioned milestone

peetucket:
...even if it is not the last timestamp

This gets around the issue where we show a status to users in Argo that is incorrect if an earlier workflow step with an earlier lifecycle is manually repeated or triggered during an object remediation.  The idea is that if an object has a current lifecycle of accessioned (current being the latest version), it should be considered accessioned regardless of timestamps.

### DIFF
--- a/lib/dor/services/status_service.rb
+++ b/lib/dor/services/status_service.rb
@@ -46,6 +46,10 @@ module Dor
 
     # @return [Hash{Symbol => Object}] including :current_version, :status_code and :status_time
     def status_info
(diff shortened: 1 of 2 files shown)


jcoyne:
Why is this on 7-2-stable and not on master?  I'd expect a change in master first and then a backport to 7-x (probably 7.3) if necessary

peetucket:
fair enough, same change on master: https://github.com/sul-dlss/dor-services/pull/648

PR #661, opened 2019-09-17
Remove method that was marked for deprecation

jcoyne:


mjgiarlo:
Blank line rubocop issue again. Dang keyboard.

lib/dor/datastreams/content_metadata_ds.rb
```diff
@@ -1,3 +1,4 @@
+
```
ndushay: d'oh!

PR #668, opened 2019-09-20
Pass the version to the active_version workflow call

jcoyne:
If you don't pass a version the workflow service will make a HTTP request to dor-services-app to get the version. This makes it very slow to open/close versions.


dor-services.gemspec
```diff
@@ -1,3 +1,4 @@
+
```
mjgiarlo: :policeman: :unamused:

PR #669, opened 2019-09-23
Deprecate StateService

jcoyne:
Argo is the only consumer, so we'll just move it there

### DIFF
--- a/lib/dor/services/state_service.rb
+++ b/lib/dor/services/state_service.rb
@@ -1,4 +1,3 @@
-
 # frozen_string_literal: true
 
 module Dor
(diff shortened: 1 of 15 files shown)


lib/dor/services/state_service.rb
```diff
@@ -7,6 +7,7 @@ class StateService
     # having version is preferred as without it, a call to
     # fedora will be made to retrieve it.
     def initialize(pid, version: nil)
+      Deprecation.warn(self, 'Dor::StateService is deprecate and will be removed in dor-services 9')
```
jmartin-sul: tiny tense typo?  "is deprecate" -> "is deprecated"

PR #673, opened 2019-12-19
Refactor workflow indexer

jcoyne:


lib/dor/indexers/workflow_indexer.rb
```diff
@@ -36,6 +33,14 @@ def to_solr
     attr_reader :document
     delegate :processes, to: :document
 
+    def wf_name
+      @wf_name ||= document.workflowId.first
+    end
+
+    def repo
+      document.repository.first
```
mjgiarlo: Can you say why one of these is memoized and not the other?

jcoyne: `repo` is only used one time.

PR #692, opened 2020-03-10
Deprecate embargoable methods

jcoyne:
## Why was this change made?

The are just indirection and we can call the EmbargoService directly

n/a

### DIFF
--- a/lib/dor/models/concerns/embargoable.rb
+++ b/lib/dor/models/concerns/embargoable.rb
@@ -17,12 +17,12 @@ module Dor
     def release_embargo(release_agent = 'unknown')
       embargo_service.release(release_agent)
     end
-    deprecation_deprecate release_embargo: 'Use the method in EmbargoService instead'
+    deprecation_deprecate release_embargo: 'this moved to dor-service-app'
 
     def release_20_pct_vis_embargo(release_agent = 'unknown')
       embargo_service.release_20_pct_vis(release_agent)
     end
-    deprecation_deprecate release_20_pct_vis_embargo: 'Use the method in EmbargoService instead'
+    deprecation_deprecate release_20_pct_vis_embargo: 'this moved to dor-service-app'
 
     def embargoed?
       embargoMetadata.status == 'embargoed'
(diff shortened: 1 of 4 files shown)


lib/dor/models/concerns/embargoable.rb
```diff
@@ -17,10 +17,12 @@ module Embargoable
     def release_embargo(release_agent = 'unknown')
       embargo_service.release(release_agent)
     end
+    deprecation_deprecate release_embargo: 'Use the method in EmbargoService instead'
```
jmartin-sul: is this advising use of the deprecated method in `EmbargoService`?  if so, why not just give the same warning as that method, to reduce warning indirection?  if that's not what's happening, apologies for misunderstanding.

lib/dor/models/concerns/embargoable.rb
```diff
@@ -17,10 +17,12 @@ module Embargoable
...
 
     def release_20_pct_vis_embargo(release_agent = 'unknown')
       embargo_service.release_20_pct_vis(release_agent)
     end
+    deprecation_deprecate release_20_pct_vis_embargo: 'Use the method in EmbargoService instead'
```
jmartin-sul: same question here

lib/dor/services/embargo_service.rb
```diff
@@ -55,6 +56,7 @@ def release_20_pct_vis(release_agent)
 
       events.add_event('embargo', release_agent, '20% Visibility Embargo released')
     end
+    deprecation_deprecate release_20_pct_vis: 'this moved to dor-service-app'
 
     def update(new_date)
```
jmartin-sul: was lack of deprecation warning on this method intentional?

ndushay: yeah - did you put the deprecation warning in the wrong place?  Or just omit one?

lib/dor/models/concerns/embargoable.rb
jcoyne: Yeah, this was 2 commits, and this message made sense with the first, but not after the second.

lib/dor/services/embargo_service.rb
jcoyne: `#update` is not being deprecated

PR #694, opened 2020-03-11
Add embargo_release_date to RightsMetadataDS

jcoyne:
## Why was this change made?
Made EmbargoMetadataDS#release_date switch to OM methods so it works in the same way as RightsMetadataDS#embargo_release_date.
As this avoids some XML parsing, it should help us migrate away from Fedora 3.

Ref https://github.com/sul-dlss/google-books/issues/399

n/a


jcoyne:
@ndushay this blocks google books work on staging and this is a gem, not a service.

PR #695, opened 2020-03-16
Embargo indexing fix

ndushay:
## Why was this change made?

To fix error in dor_indexing_app tests when updating to dor-services 9.2.0 https://github.com/sul-dlss/dor_indexing_app/pull/373

Note that 
- there was no test for the code change that broke it
- it may be unrealistic for there to be an embargo without a release date in our actual data

na

### DIFF
--- a/lib/dor/datastreams/embargo_metadata_ds.rb
+++ b/lib/dor/datastreams/embargo_metadata_ds.rb
@@ -36,7 +36,7 @@ module Dor
     def to_solr(solr_doc = {}, *args)
       solr_doc = super
       rd20 = twenty_pct_release_date
-      ::Solrizer.insert_field(solr_doc, 'embargo_release', release_date.first.utc.strftime('%FT%TZ'), :dateable) unless release_date.blank?
+      ::Solrizer.insert_field(solr_doc, 'embargo_release', release_date.first.utc.strftime('%FT%TZ'), :dateable) unless release_date.first.blank?
       ::Solrizer.insert_field(solr_doc, 'twenty_pct_visibility_release', rd20.utc.strftime('%FT%TZ'), :dateable) unless rd20.blank?
       solr_doc
     end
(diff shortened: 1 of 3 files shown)


ndushay:
I've confirmed that this fixes the problem in dor_indexing_app when using dor-services 9.2.0.

lib/dor/datastreams/embargo_metadata_ds.rb
```diff
@@ -36,7 +36,7 @@ def self.xml_template
     def to_solr(solr_doc = {}, *args)
       solr_doc = super
       rd20 = twenty_pct_release_date
-      ::Solrizer.insert_field(solr_doc, 'embargo_release', release_date.first.utc.strftime('%FT%TZ'), :dateable) unless release_date.blank?
+      ::Solrizer.insert_field(solr_doc, 'embargo_release', release_date.first.utc.strftime('%FT%TZ'), :dateable) unless release_date&.first.blank?
```
jcoyne: I don't think this is correct.  if `release_date` returns `[nil]` then this isn't going to prevent the error.  Can you write a test?

jcoyne:
```
Dor::Item.new.embargoMetadata.release_date
=> [nil]
```

jcoyne: Oh, actually it would prevent the error, but the `&.` is unnecessary.

ndushay: release_date returns `[nil]` and I started to write a test but ... isn't this just for the other code with empty data?  would we ever an empty release date?

jcoyne: 🤷‍♂ Reviewer didn't ask me to.

jcoyne: I'm suggesting `[nil].first.blank?`

ndushay: yes, that is what is broken and what this fixes.  I can make a test but I'm asking if that is realistic in a fedora embargo datastream ...

jcoyne: I'm saying get rid of the `&` as that doesn't do anything

ndushay: @jcoyne and fixed again again.

PR #696, opened 2020-03-20
Remove outdated and unused administrative tags code

mjgiarlo:
Connects to sul-dlss/dor-services-app#679

This will break DSA. Before merging this, we might want to extract the tag service into DSA, as a way to continue testing that continual migration functions as expected.

## Why was this change made?

To remove dead code.

No.


mjgiarlo:
Blocked by https://github.com/sul-dlss/dor-services-app/pull/758

PR #704, opened 2020-05-20
Set the agreement class correctly

jcoyne:
## Why was this change made?

If it's not agreement, then previously set agreements are not removed.

See https://github.com/samvera/active_fedora/blob/8.x-stable/lib/active_fedora/associations/belongs_to_association.rb#L91


peetucket:
Cool.  So I follow on steps would be new release of dor-services, I could test locally with the new release, then bump in Argo, and then new argo release?

PR #713, opened 2020-08-06
Allow new controlled digital lending rights to be set

### LINKED ISSUE #712: Add new CDL option (controlled digital lending) as an option for object rights
See sul-dlss/argo#2200 for full details.  Basically we need a new option for rights and the ability to generate this option in rightsMetadata.  It will be available for selection in Argo.

New option to be added here: https://github.com/sul-dlss/dor-services/blob/master/lib/dor/datastreams/rights_metadata_ds.rb#L62-L75 

And we need the ability to generate the correct rights block

peetucket:
## Why was this change made?

Fixes #712 

## How was this change tested?

New tests


spec/datastreams/content_metadata_ds_spec.rb
```diff
@@ -248,14 +248,14 @@
   end
 
   describe '#contentType=' do
-    let(:contentMetadata) { described_class.new }
+    let(:content_metadata) { described_class.new }
```
peetucket, line 251: didn't touch this code, only to make rubocop happy

lib/dor/datastreams/rights_metadata_ds.rb
```diff
@@ -121,6 +126,15 @@ def self.upd_rights_xml_for_rights_type(rights_xml, rights_type)
           loc_node = Nokogiri::XML::Node.new('location', rights_xml)
           loc_node.content = rights_type.split(':').last
           machine_node.add_child(loc_node)
+        elsif rights_type.start_with?('cdl')
+          cdl_node = Nokogiri::XML::Node.new('cdl', rights_xml)
+          if rights_type.end_with?('-nd')
```
jcoyne, line 131: Are there any cases where this is not always true?

spec/datastreams/rights_metadata_spec.rb
```diff
@@ -240,6 +240,32 @@
...
+            <human type="copyright">Courtesy of the Revs Institute for Automotive Research. All rights reserved unless otherwise indicated.</human>
+          </copyright>
+          <access type="discover">
+            <machine>
+              <world/>
+            </machine>
+          </access>
+          <access type="read">
+            <machine>
+              <cdl>
```
jcoyne, line 256: Shouln't `<cdl />`  be separate from and in addition to a `<group rule="no-download">stanford</group>`. that way when `<cdl />` is removed at the end of the term, we just remove the one node?

spec/datastreams/rights_metadata_spec.rb
```diff
@@ -265,6 +291,12 @@
       expect(rights_ng_xml).to be_equivalent_to stanford_no_download_rights_xml
     end
 
+    it 'has the expected rights xml when read rights are set to controlled digital lending with the no-download rule' do
+      rights_ng_xml = Nokogiri::XML(original_rights_xml)
+      described_class.upd_rights_xml_for_rights_type(rights_ng_xml, 'cdl-nd')
```
jcoyne, line 296: shouldn't this be `cdl-stanford-nd`?

lib/dor/datastreams/rights_metadata_ds.rb
```diff
@@ -59,11 +59,14 @@ def self.xml_template
       end.doc
     end
 
+    # NOTE: 8/6/2020 In the future, 'cdl' could be a valid option in the list below.  It will generate a <cdl> node
```
jcoyne, line 62: Let's not build this until we know we'll need it.

spec/datastreams/rights_metadata_spec.rb
peetucket: yes, i'll fix that

lib/dor/datastreams/rights_metadata_ds.rb
peetucket: if we only have cdl-stanford-nd, that is true that we don't this, and can add it back later if we need to support that distinction

spec/datastreams/rights_metadata_spec.rb
peetucket: the way the rights were specified by chris beer here https://docs.google.com/document/d/14JPhnBefJWcKZ0uFePZ1yn555aZwBoBigrE9-bsLbmg under the "straw proposal" section, he had the no-download node within the <cdl> node, like this:

```
<cdl>
  <group rule="no-download">stanford</group>
</cdl>
```

I'm not sure if that is important to drive the behavior.  But I believe removing the <cdl> node at the end will take that embedded node with it, right?

lib/dor/datastreams/rights_metadata_ds.rb
peetucket: ok

spec/datastreams/rights_metadata_spec.rb
jcoyne: Yes and since that would take that away, don't we want something to fall back to?

jcoyne: Also it would be great if this aligned with the description in https://github.com/sul-dlss/argo/issues/2200

spec/datastreams/content_metadata_ds_spec.rb
jcoyne: Lets take care of that separately then: https://github.com/sul-dlss/dor-services/pull/714

spec/datastreams/rights_metadata_spec.rb
peetucket: Good call, I fixed the ticket description.   We may need to run this by Chris again to be sure we understand the intent.  I think the idea is that when this block is taken away, a new read block would be written to make it citation only again.

spec/datastreams/content_metadata_ds_spec.rb
peetucket: Hmm... I merged that PR, and then rebase on this branch, but now I get various other rubocop violations in other parts of the code.

jcoyne: You have to rebase and then fix them here or this will cause those errors on master

spec/datastreams/rights_metadata_spec.rb
peetucket: Checked with Chris and the ticket is accurate.

spec/datastreams/content_metadata_ds_spec.rb
peetucket: Backed this change out of this PR and created a new PR for this and some other rubocop fixes I saw in master branch: #715

peetucket: I'm confused.  I've not updated my local rubocop gem for this project but it looks like I'm now using the same gem as travis, but the errors I am getting (none now) do not align with theirs.

peetucket: Ah, ok, I needed to update other rubocop related gems.  Now I get the same errors as travis...back to another PR i guess.

peetucket: Ok, new rubocop PR to fix failures on master branch, can rebase to them here too: #716

peetucket:
Did another rubocop auto gen because the tests added another example in the same spot.

PR #715, opened 2020-08-17
new rubocop fixes

peetucket:
## Why was this change made?

~~Rubocop failing locally for me now on master branch~~

Never mind, rubocop is working fine on master when i updated my local version of the rubocop gem.  So this is not needed to make rubocop happy,

## How was this change tested?

Run the test suite locally, no rubocop or test failures

## Which documentation and/or configurations were updated?

None


peetucket:
OK, my local rubocop results seems to differ from travis and I'm not sure why.

peetucket:
Oh, hmm, may have a different version of the gem.

PR #717, opened 2020-08-26
bump dor-rights-auth gem

peetucket:
## Why was this change made?

The new dor-rights-auth gem has some updated CDL parsing code.

### DIFF
--- a/dor-services.gemspec
+++ b/dor-services.gemspec
@@ -32,7 +32,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'solrizer', '~> 3.0'
 
   # Stanford dependencies
-  s.add_dependency 'dor-rights-auth', '~> 1.0', '>= 1.2.0'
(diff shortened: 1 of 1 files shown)


dor-services.gemspec
```diff
@@ -32,7 +32,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'solrizer', '~> 3.0'
 
   # Stanford dependencies
-  s.add_dependency 'dor-rights-auth', '~> 1.0', '>= 1.2.0'
+  s.add_dependency 'dor-rights-auth', '~> 1.0', '>= 1.5.0'
```
jcoyne, line 35: Is there is anything in dor-services that requires the new methods we introduced?  I think the dependency lies in DSA.

peetucket: that's right, the dependency is in DSA, but DSA doesn't specifically require dor-rights-auth right now... so an alternate approach is to go to DSA and specifically ask for the correct version of dor-rights-auth there

PR #719, opened 2020-09-29
Make use_statement and copyright writeable

jcoyne:
When the source xml has no existing nodes

## Why was this change made?
 Fixes:
```
item = Dor::Item.new
item.rightsMetadata.content = "<?xml version=\"1.0\"?>\n<rightsMetadata>\n <access type
=\"discover\">\n <machine>\n <none/>\n </machine>\n </access>\n <access type=\"r
ead\">\n <machine>\n <none/>\n </machine>\n </access>\n</rightsMetadata>\n"
item.rightsMetadata.copyright = 'foo'
Traceback (most recent call last):
2: from (irb):6
1: from (eval):1:in `block (2 levels) in insert_from_template'
NoMethodError (undefined method `human' for #<Dor::RightsMetadataDS:0x000000000326d420>)
```

See https://app.honeybadger.io/projects/50568/faults/68130941


.rubocop_todo.yml
```diff
@@ -56,7 +56,7 @@ Lint/Void:
 # Offense count: 27
 # Configuration parameters: IgnoredMethods.
 Metrics/AbcSize:
-  Max: 73
+  Max: 74
```
jcoyne, line 59: Needed to regenerate this because rubocop changed how it calculated AbcSize and we were over the threshold.

justinlittman (approved):

